PR #36, opened 2019-04-12
Issue 32

sharkovadarya:


src/test/kotlin/com/jetbrains/snakecharm/SnakemakeResolveTest.kt
```diff
@@ -8,20 +8,6 @@ import com.jetbrains.python.psi.PyFunction
 import com.jetbrains.snakecharm.fixtures.SnakemakeResolveTestCase
```
iromeo: Please remove "unused imports" that appeared due to this cleanup. It is good practice "do not commit yellow code", so it is recommended to resolve code insight warning before commit
image

src/test/kotlin/com/jetbrains/snakecharm/fixtures/SnakemakeResolveTestCase.kt
```diff
@@ -1,21 +1,12 @@
...
     companion object {
-        fun <T : PsiElement> assertResolveResult(element: PsiElement?,
-                                                 aClass: Class<T>,
-                                                 name: String): T {
-            return assertResolveResult(element, aClass, name, null)
-        }
-
         fun <T : PsiElement> assertResolveResult(element: PsiElement?,
                                                  aClass: Class<T>,
                                                  name: String,
```
iromeo: According to Kotline code style guidelines (https://kotlinlang.org/docs/reference/coding-conventions.html):
> As a general rule, avoid horizontal alignment of any kind. 

also see function formatting conventions https://kotlinlang.org/docs/reference/coding-conventions.html#function-formatting

So here better is to format like
```
        fun <T : PsiElement> assertResolveResult(
                element: PsiElement?,
                aClass: Class<T>,
                name: String,
                containingFilePath: String?
        ): T {
```

src/test/kotlin/com/jetbrains/snakecharm/fixtures/SnakemakeResolveTestCase.kt
```diff
@@ -0,0 +1,134 @@
...
+
+    protected fun <T : PsiElement> assertResolvesTo(aClass: Class<T>,
+                                                    name: String,
+                                                    containingFilePath: String?): T {
+        val element: PsiElement?
+        try {
+            element = doResolve()
+        } catch (e: Exception) {
+            throw RuntimeException(e)
+        }
```
iromeo: In Kotlin all exceptions are checked, we don't need to re-throw exception here, see https://kotlinlang.org/docs/reference/java-interop.html#checked-exceptions. Please clean up the code, it also could be converted to expression body e.g. `fun foo() = boo(...)`

src/test/kotlin/com/jetbrains/snakecharm/fixtures/SnakemakeResolveTestCase.kt
```diff
@@ -33,35 +24,17 @@ abstract class SnakemakeResolveTestCase : SnakemakeTestCase() {
...
-        return result.get()
     }
 
-    protected fun <T : PsiElement> assertResolvesTo(aClass: Class<T>, name: String): T {
-        return assertResolvesTo(aClass, name, null)
+    private fun doResolve(): PsiElement? {
+        val ref = fixture?.getReferenceAtCaretPosition("resolve/" +
+                getTestName(false) + ".smk")
+        return ref?.resolve()
     }
```
iromeo: Please
1) inline `ref` variable 
2) convert function to expression body
3) replace string concatenation with Kotlin style, e.g. `foo/$arg/boo/${foo(arg2)}/doo.smk`

src/test/kotlin/com/jetbrains/snakecharm/fixtures/SnakemakeResolveTestCase.kt
```diff
@@ -0,0 +1,134 @@
...
+        protected const val MARKER = "<ref>"
+
+        fun <T : PsiElement> assertResolveResult(element: PsiElement?,
+                                                 aClass: Class<T>,
+                                                 name: String): T {
+            return assertResolveResult(element, aClass, name, null)
+        }
+
+        fun <T : PsiElement> assertResolveResult(element: PsiElement?,
+                                                 aClass: Class<T>,
```
iromeo: To my mind in methods like JUnit assert* it is good to explicitly specify which values are expected, which parameters are actual. So here I'd rather name variables
```
        fun <T : PsiElement> assertResolveResult(
                actualElement: PsiElement?,
                expectedClass: Class<T>,
                expectedElementName: String,
                expectedContainingFilePath: String?
        ): T {
```

E.g. similar to 

```
  public static void assertEquals(String expected, String actual) {
    Assert.assertEquals(expected, actual);
  }
```

src/test/kotlin/com/jetbrains/snakecharm/SnakemakeResolveTest.kt
```diff
@@ -31,7 +31,10 @@ class SnakemakeResolveTest : SnakemakeResolveTestCase() {
     }
 
     fun testExpandTopLevel() {
-        //val result = resolve()
+        assertResolvesTo(PyFunction::class.java, "expand")
+    }
+
+    fun testExpandNested() {
         assertResolvesTo(PyFunction::class.java, "expand")
     }
```
iromeo: Please add more tests, like you have for completion. E.g.
* expand could be resolved out of `rule` block scope, just some top-level python code in file
* expand won't be resolved in not snakemake *.py file

src/main/kotlin/com/jetbrains/snakecharm/lang/psi/resolve/SnakemakeReferenceResolveProvider.kt
```diff
@@ -1,26 +1,24 @@
...
 
-    override fun resolveName(element: PyQualifiedExpression, context: TypeEvalContext): MutableList<RatedResolveResult> {
+    override fun resolveName(element: PyQualifiedExpression, context: TypeEvalContext): List<RatedResolveResult> {
         val facade = PyPsiFacade.getInstance(element.project)
-        val fullyQualifiedName = namesToFQN[element.name] ?: return mutableListOf()
-        val qName = QualifiedName.fromDottedString(fullyQualifiedName)
 
-        val resolveContext = facade.createResolveContextFromFoothold(element)
-        val resolvedModules = facade.resolveQualifiedName(qName, resolveContext)
+        val fullyQualifiedName = namesToFQN[element.name] ?: return emptyList()
```
iromeo: I assume that this resolve is provided for any *.py file even not related to snakemake files

src/test/kotlin/com/jetbrains/snakecharm/fixtures/SnakemakeResolveTestCase.kt
iromeo: done

src/test/kotlin/com/jetbrains/snakecharm/fixtures/SnakemakeResolveTestCase.kt
```diff
@@ -33,35 +24,17 @@ abstract class SnakemakeResolveTestCase : SnakemakeTestCase() {
...
-                Runnable { result.set(assertResolvesTo(aClass, name, null)) }
-        )
-
-        return result.get()
     }
 
-    protected fun <T : PsiElement> assertResolvesTo(aClass: Class<T>, name: String): T {
-        return assertResolvesTo(aClass, name, null)
+    private fun doResolve(): PsiElement? {
+        val ref = fixture?.getReferenceAtCaretPosition("resolve/" +
```
iromeo: Here is ok to expect that `fixture != null` because it is true if test setup was called. Also let's make 2 cases different: unresolved reference case (a bug in resolve implementation), no reference at caret (a bug in test data). I've already fixed that, no action required.

src/test/kotlin/com/jetbrains/snakecharm/fixtures/SnakemakeResolveTestCase.kt
iromeo: Done

src/test/kotlin/com/jetbrains/snakecharm/fixtures/SnakemakeResolveTestCase.kt
iromeo: done

src/test/kotlin/com/jetbrains/snakecharm/fixtures/SnakemakeResolveTestCase.kt
iromeo: Done

src/main/kotlin/com/jetbrains/snakecharm/lang/psi/resolve/SnakemakeReferenceResolveProvider.kt
iromeo: Done

src/test/kotlin/com/jetbrains/snakecharm/SnakemakeResolveTest.kt
iromeo: Done

PR #51, opened 2019-05-20
Test task 1: resolve and completion for rule names ('rules' object)

sharkovadarya:


src/main/kotlin/com/jetbrains/snakecharm/lang/psi/types/SnakemakeRulesType.kt
```diff
@@ -0,0 +1,65 @@
...
+        }
+
+        val rules = getRules(location)
+        val result = ArrayList<Any>()
+        rules.forEach { result.add(LookupElementBuilder.create(it.first)) }
+
+        return ArrayUtil.toObjectArray(result)
+    }
+
+    override fun assertValid(message: String?) { }
```
iromeo: Why the method isn't implemented? Is type always valid?

PR #52, opened 2019-05-20
First problem solution

nekit-000000:


iromeo:
* Doesn't work properly
* No tests provided

https://github.com/JetBrains-Research/snakecharm/pull/51 was accepted instead of this.

Issue #43, opened 2019-05-11
'shell' section: resolve & completion for `params` args


iromeo:
'shell' section: resolve & completion for `params` args

```
rule call_peaks_span_all:
    input:
        bam="cleaned_all/pooled.bam",
        span=rules.download_span.output,
        chrom_sizes=rules.download_chrom_sizes.output
    output: "cleaned_all_peaks/span/pooled_{bin, [0-9]+}.span"
    params:
        outdir=lambda wildcards, output: os.path.dirname(str(output)),
        xmx=lambda wildcards: str(800 // int(wildcards.bin))
    threads: 8
    shell: 'java -Xmx{params.xmx}G  -jar {input.span} analyze -t {input.bam} --workdir {params.outdir} --fragment 0 ' \
           '--bin {wildcards.bin} --cs {input.chrom_sizes} --threads {threads} --model {output} --keep-dup true --debug; '\
           '--foo {params.}'
```

* Auto-complete keyword args names after `params.` in shell section. E.g.
![2019-05-11_18-20-29](https://user-images.githubusercontent.com/72933/57572609-79590580-7425-11e9-967a-b4049cb3ac96.png)

* Resolve keyword args names, e.g. `xmx` in `{params.xmx}` into `params:` section

iromeo:
Merged in master, feature throws an exception:
```
rule foo:
    params: 1, ff=1, bb=2
    shell: "ffo {params.} boo"
```

```
kotlin.KotlinNullPointerException
	at com.jetbrains.snakecharm.lang.psi.SMKParamsReference.getVariants(SMKParamsReference.kt:34)
	at com.intellij.codeInsight.completion.CompletionData.completeReference(CompletionData.java:274)
	at com.intellij.codeInsight.completion.CompletionVariant.addReferenceCompletions(CompletionVariant.java:137)
	at com.intellij.codeInsight.completion.CompletionData.completeReference(CompletionData.java:82)
	at com.intellij.codeInsight.completion.LegacyCompletionContributor.lambda$completeReference$0(LegacyCompletionContributor.java:71)
	at com.intellij.codeInsight.completion.LegacyCompletionContributor.processReference(LegacyCompletionContributor.java:125)
	at com.intellij.codeInsight.completion.LegacyCompletionContributor.processReferences(LegacyCompletionContributor.java:108)
	at com.intellij.codeInsight.completion.LegacyCompletionContributor.completeReference(LegacyCompletionContributor.java:69)
[...]

PR #54, opened 2019-05-24
Issue 43

sharkovadarya:
'shell' section: resolve & completion for `params` args https://github.com/JetBrains-Research/snakecharm/issues/43


src/main/kotlin/com/jetbrains/snakecharm/codeInsight/SnakemakeParamsInShellReferenceContributor.kt
```diff
@@ -0,0 +1,48 @@
...
+import com.jetbrains.python.psi.PyStringLiteralExpression
+import com.jetbrains.snakecharm.lang.SnakemakeLanguageDialect
+import com.jetbrains.snakecharm.lang.psi.SMKParamsReference
+import com.jetbrains.snakecharm.lang.psi.SMKRuleParameterListStatement
+import java.util.regex.Pattern
+
+class SnakemakeParamsInShellReferenceContributor : PsiReferenceContributor() {
+    override fun registerReferenceProviders(registrar: PsiReferenceRegistrar) {
+        registrar.registerReferenceProvider(
+                PlatformPatterns.psiElement(),
```
iromeo: Could we use a narrower filter, e.g only for python string-like psi elements?

src/main/kotlin/com/jetbrains/snakecharm/codeInsight/SnakemakeParamsInShellReferenceContributor.kt
```diff
@@ -0,0 +1,48 @@
...
+                object : PsiReferenceProvider() {
+                    private val paramsPattern = Pattern.compile("\\{params\\.(.+?)}")
```
iromeo: Do we want to accept any suffix after `params.` e.g. `foo.boo` in `params.foo.boo`? Seems we expect only identifier after dot.

src/main/kotlin/com/jetbrains/snakecharm/lang/psi/SMKParamsReference.kt
```diff
@@ -0,0 +1,54 @@
...
+        val paramsSection = getParamsSection() ?: return emptyArray()
+        val variants = ArrayList<LookupElement>()
+        paramsSection.argumentList?.arguments?.forEach {
+            variants.add(LookupElementBuilder.create(it.name!!))
+        }
+        return variants.toTypedArray()
+    }
+
+    private fun getParamsSection(): SMKRuleParameterListStatement? {
+        val rule = this.element.parentOfType<SMKRule>() ?: return null
```
iromeo: Better to improve SMKRule psi to access section element with desired name

src/main/kotlin/com/jetbrains/snakecharm/lang/psi/SMKParamsReference.kt
```diff
@@ -0,0 +1,54 @@
...
+    override fun resolve(): PsiElement? {
+        val resolveResults = multiResolve(false)
+        return if (resolveResults.size == 1) resolveResults[0].element else null
+    }
+
+    override fun getVariants(): Array<Any> {
+        val paramsSection = getParamsSection() ?: return emptyArray()
+        val variants = ArrayList<LookupElement>()
+        paramsSection.argumentList?.arguments?.forEach {
+            variants.add(LookupElementBuilder.create(it.name!!))
```
iromeo: Icon is missing =(

src/main/kotlin/com/jetbrains/snakecharm/lang/psi/SMKParamsReference.kt
```diff
@@ -0,0 +1,54 @@
...
+import com.jetbrains.python.psi.resolve.RatedResolveResult
+import java.util.*
+
+
+class SMKParamsReference(
+        element: PsiElement, textRange: TextRange
+) : PsiReferenceBase<PsiElement>(element, textRange), PsiPolyVariantReference {
+    private val key: String = element.text.substring(textRange.startOffset, textRange.endOffset)
+
+    override fun multiResolve(incompleteCode: Boolean): Array<ResolveResult> {
```
iromeo: What is the difference with `resolve` method here? Does this method add any value? Does it really do "multi-resolve" ?

src/main/kotlin/com/jetbrains/snakecharm/lang/psi/SMKParamsReference.kt
```diff
@@ -0,0 +1,54 @@
...
+        val paramsSection = getParamsSection() ?: return emptyArray()
+        paramsSection.argumentList?.arguments?.forEach {
```
iromeo: Code duplication here and in get variants method. Please extract method which collects keyword arguments.

src/main/kotlin/com/jetbrains/snakecharm/codeInsight/SnakemakeParamsInShellReferenceContributor.kt
sharkovadarya: I don't quite get what's the desired behavior here, probably because I've been unable to find an example of nested parameters declaration syntax (how would I go about declaring `foo.boo` in the `params` section?). As of now, any attempt at writing `params.foo.boo` when only `foo` is defined results in an unresolved reference error. What should the user get instead? I could modify the regex to not match `params.foo.`, then there would be no error, but I'm not sure if that's what's expected here. I don't really know what happens in case if `params.foo.boo` should actually get resolved because I'm unable to declare such a construction.

iromeo: I think, that we cannot process nested params at the moment and thus I expect to match only one identifier after `params.`, not any complex expression.

src/test/resources/com/jetbrains/snakecharm/cucumber/resolve/snakemake_params_in_shell.feature
```diff
@@ -31,9 +31,14 @@ Feature: Resolve for params in shell section
...
           xmx=os
-        shell: "command {params.xmx.path}"
+        shell: "command {params.xmx<suffix>}"
       """
       When I put the caret after {params.xm
       Then reference should resolve to "xmx" in "foo.smk"
+      Examples:
+        | suffix   |
+        | .path    |
+        | [path]   |
```
iromeo: @sharkovadarya Regexp in reference impl doesn't work for 2nd and 3rd cases. Please update regexp to accept identifier (e.g. a sequence of word letters) and then some non-word letter.

src/test/resources/com/jetbrains/snakecharm/cucumber/resolve/snakemake_params_in_shell.feature
```diff
@@ -20,6 +21,7 @@ Feature: Resolve for params in shell section
       | ptn         | text          | symbol_name | file         |
       | {params.out | outdir        | outdir      | foo.smk      |
       | {params.xm  | xmx           | xmx         | foo.smk      |
+      | {params.fi  | file1         | file1       | foo.smk      |
```
iromeo: @sharkovadarya Regexp still doesn't support all valid identifiers.

src/main/kotlin/com/jetbrains/snakecharm/codeInsight/SnakemakeParamsInShellReferenceContributor.kt
```diff
@@ -0,0 +1,48 @@
...
+                            element: PsiElement,
+                            context: ProcessingContext
+                    ): Array<PsiReference> {
+                        if (element.containingFile.language != SnakemakeLanguageDialect) {
+                            return emptyArray()
+                        }
+
+                        val paramReferences = mutableListOf<PsiReference>()
+
+                        val isShellCommand = element
```
iromeo: See `RuleSectionKeywordsProvider`, you can move language check and `inside SMKRuleParameterListStatement` check into pattern used in `registrar.registerReferenceProvider`

sharkovadarya: I can remove the `is PyStringLiteral` check but I haven't been able to come up with the correct pattern for checking if we are inside the _shell_ section, not just any rule section: the element we're looking at is a string literal, its parent is a `PyArgumentList`, and its parent is `SMKRuleParameterListStatement`, whose name, according to its `getName()` method, is "shell" but who does not match the `PlatformPatterns.psiElement(SMKRuleParameterListStatement::class.java).withName("shell")` pattern somehow. So, yes, I can modify the pattern used in `registrar.registerReferenceProvider`, but I still don't see how I can move the `isShellCommand` check to the pattern because the pattern that, I think, should work, doesn't work, and I don't see how to form this pattern in another way.

iromeo: I meant only check that element is inside required parent and in file with proper language, not section name

sharkovadarya: That means autocompletion/resolve for params would be available not only in shell section, but in all other rule sections as well. Should I add corresponding tests?

src/test/resources/com/jetbrains/snakecharm/cucumber/completion/snakemake_params_in_shell.feature
```diff
@@ -16,23 +16,5 @@ Feature: Completion for params in shell section
     When I put the caret after {params.
     And I invoke autocompletion popup
     Then completion list should contain:
-      | outdir      |
-      | xmx         |
-
-  Scenario: Not completed in wrapper section
```
iromeo: please put back the test also

src/main/kotlin/com/jetbrains/snakecharm/codeInsight/SnakemakeParamsInShellReferenceContributor.kt
sharkovadarya: I fixed the pattern but the safe call in `.getParentOfType(element, SMKRuleParameterListStatement::class.java)?.section` still can't be removed, compilation error.

src/main/kotlin/com/jetbrains/snakecharm/codeInsight/SnakemakeParamsInShellReferenceContributor.kt
```diff
@@ -26,12 +26,17 @@ class SnakemakeParamsInShellReferenceContributor : PsiReferenceContributor() {
...
                         val paramReferences = mutableListOf<PsiReference>()
-
                         val paramsMatcher = paramsPattern.matcher(element.text)
 
-                        while (paramsMatcher.find()) {
-                            paramReferences.add(SMKParamsReference(element,
-                                    TextRange(paramsMatcher.start(1), paramsMatcher.end(1))))
+                        val isShellCommand = PsiTreeUtil
+                                .getParentOfType(element, SMKRuleParameterListStatement::class.java)
+                                ?.section
```
iromeo: Could `.getParentOfType(element, SMKRuleParameterListStatement::class.java)` be null here?

iromeo: the question is discussed in another comment

sharkovadarya: No it could not, but I have tried removing this safe call and code didn't compile, as I have mentioned above.

src/main/kotlin/com/jetbrains/snakecharm/codeInsight/SnakemakeParamsInShellReferenceContributor.kt
iromeo: What kind of compilation error?

iromeo:
Was closed by a mistake

Issue #12, opened 2019-01-11
Completion for `shadow` level values


iromeo:
E.g. `full`, `minimal`, `shallow`, .... See https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html?highlight=shadow#shadow-rules

iromeo:
@nekit-000000 Please attach here a screenshot which demonstrates how the feature works (e.g. show auto-completion list)

nekit-000000:
![shadowSettingsCompletion](https://user-images.githubusercontent.com/25721619/60179820-357c5d00-9827-11e9-92fe-09258895cff0.png)

iromeo:
Feature will be available in 0.4.x plugin verison

PR #63, opened 2019-06-26
'shadow' settings completion added with tests

nekit-000000:
#12


iromeo:
Please use relevant commit messages, explaining the idea of the change. At the moment you've pushed 2 commits with exact same commit messages although different content.

PR #64, opened 2019-06-26
Feature/shadow settings inspection

nekit-000000:


src/main/kotlin/com/jetbrains/snakecharm/inspections/SmkShadowSettingsInspection.kt
```diff
@@ -0,0 +1,37 @@
...
+    ) = object : SnakemakeInspectionVisitor(holder, session) {
+
+        override fun visitSMKRuleParameterListStatement(st: SMKRuleParameterListStatement) {
+            if (!SnakemakeLanguageDialect.isInsideSmkFile(st) ||
+                    st.name != SMKRuleParameterListStatement.SHADOW) {
+                return
+            }
+
+            val stringLiteral = st.argumentList?.arguments?.find {
+                it is PyStringLiteralExpression
```
iromeo: Does snakemake allow more that one arg here in arg list for "shadow" section? If not, than we could also highlight it as a problem

nekit-000000: Only one argument after **shadow** section is permitted. If you try to run this script:
```
rule NAME:
    shadow: "full", "shallow"
```
you will get an error: 
_shadow() takes 2 positional arguments but 3 were given_

src/main/kotlin/com/jetbrains/snakecharm/inspections/SmkShadowSettingsInspection.kt
```diff
@@ -21,11 +21,10 @@ class SmkShadowSettingsInspection : SnakemakeInspection()  {
                 return
             }
 
-            val stringLiteral = st.argumentList?.arguments?.find {
-                it is PyStringLiteralExpression
-            } as PyStringLiteralExpression?
+            val stringLiteral = st.argumentList?.arguments?.get(0)
```
iromeo: Please use `.first` instead of `.get(0)`

src/main/kotlin/com/jetbrains/snakecharm/inspections/SmkShadowSettingsInspection.kt
```diff
@@ -21,12 +21,12 @@ class SmkShadowSettingsInspection : SnakemakeInspection()  {
                 return
             }
 
-            val stringLiteral = st.argumentList?.arguments?.get(0)
+            val argument = if (st.argumentList?.arguments?.size != 0) st.argumentList!!.arguments[0] else null
```
iromeo: There are couple of useful methods like `st.argumentList?.arguments?.isEmpty()`, `isNotEmpty`, `isNullOrEmpty`.

Also is better to use `st.argumentList?.arguments?.firstOrNull()` it is much  shorter

src/main/kotlin/com/jetbrains/snakecharm/inspections/SmkShadowSettingsInspection.kt
```diff
@@ -21,12 +21,12 @@ class SmkShadowSettingsInspection : SnakemakeInspection()  {
...
 
-            if (stringLiteral != null &&
-                    stringLiteral is PyStringLiteralExpression &&
-                    stringLiteral.stringValue !in ShadowSectionSettingsProvider.SHADOW_SETTINGS) {
-                registerProblem(stringLiteral.originalElement,
+            if (argument != null &&
+                    argument is PyStringLiteralExpression &&
```
iromeo: I assume that it is false that `null is PyStringLiteralExpression` because `PyStringLiteralExpression` is not-null type, so you could simplify the condition, but please check it before.

iromeo:
Issue #65 done

Issue #31, opened 2019-02-06
Rule docstrings trigger a syntax error


wjv:
Snakemake explicitly supports docstrings (bare strings as the first element in a rule, analogous to docstrings in Python). See:

https://bitbucket.org/snakemake/snakemake/issues/398/making-the-code-and-docstring-of-each-rule

Currently SnakeCharm marks a docstring as a syntax error (“Rule parameter identifier is expected”).

wjv:
Let me amend this: It _sometimes_ triggers a syntax error. I am trying to isolate when exactly this happens.

iromeo:
Thanks 

Code example:
```
rule a:
    """ the docstring """
    input: ...
    ...
```

iromeo:
Issue demonstration + how *.py compilation looks like
![image](https://user-images.githubusercontent.com/72933/60290738-10b8e000-9922-11e9-8623-82db1ea9e22a.png)

iromeo:
P.S: also let's ensure that fix also resolves errors in code like
```
rule a:
    # boooo
    """ the docstring """
    input: "foo"
```

iromeo:
Update:
Also valid snakemake syntax:
```
rule a:
    " snakemake also treats this as doc string "
    input: "foo"
```

iromeo:
Fixed, will be available in next major update (0.4.x)

PR #67, opened 2019-06-30
#31 Rule docstrings trigger a syntax error

nekit-000000:
#31


testData/psi/Docstrings.smk
```diff
@@ -0,0 +1,4 @@
+rule NAME: """Docstring 1"""
```
iromeo: I think several docstrings is not the best example and probably a side-effect of  snakemake implementation. So I suggest use several tests for this feature. E.g. 2 real life use-cases, e.g. 1) single line doc string 2) multiline doc string. And a couple of syntetic usecases e.g. 3) for non """ quotes and 4) for some crazy things like several docstrings, docstrings starting on same line as rule, e.tc.

testData/psi/MultiLineDocstrings.smk
```diff
@@ -0,0 +1,5 @@
+rule NAME:
+    """Docstring 1"""
```
iromeo: Multiline doc string is
```
rule Name:
    """ line 1
    line2
    """
```

PR #74, opened 2019-07-02
Issue 71

sharkovadarya:


src/main/resources/SnakemakeBundle.properties
```diff
@@ -22,3 +22,7 @@ INSP.NAME.shadow.settings=Shadow must either be 'shallow', 'full', 'minimal', or
 
 # SnakemakeMultipleShadowSettingsInspection
 INSP.NAME.shadow.multiple.settings=Only one argument after shadow section is permitted.
+
+# SmkSectionRedeclarationInspectoin
+INSP.NAME.section.redeclaration=Section declaration above overshadows this declaration
```
iromeo: According to the Oxford dictionary "overshadow - to make smb/smth seem less important, or successful", is it a right verb for this case? Maybe just "overrides"?

src/test/kotlin/com/jetbrains/snakecharm/inspections/SmkSectionRedeclarationInspectionTest.kt
```diff
@@ -0,0 +1,21 @@
+package com.jetbrains.snakecharm.inspections
+
+import com.intellij.codeInspection.LocalInspectionTool
+import com.jetbrains.snakecharm.fixtures.SnakemakeInspectionTestCase
+
+class SmkSectionRedeclarationInspectionTest : SnakemakeInspectionTestCase() {
+    override val inspectionClass: Class<out LocalInspectionTool>
+        get() = SmkSectionRedeclarationInspection::class.java
```
iromeo: Here simpler is `override val inspectionClass = SmkSectionRedeclarationInspection::class.java`. But please re-write this using cucumber features.

Issue #72, opened 2019-07-01
Keyword arguments duplication


sharkovadarya:
Example:
 
    params: a="a1", a="a2"

results in the following syntax error: `keyword argument repeated`.

iromeo:
![image](https://user-images.githubusercontent.com/72933/60678521-6b57bc00-9e8d-11e9-8256-d9d00ddd206c.png)

iromeo:
Please leave highlighting only on keyword arg name, because no errors in arg value. Otherwise looks not very clear, like some missing comma or other syntax error

![image](https://user-images.githubusercontent.com/72933/61997572-1b46d080-b0ac-11e9-8d04-9b8aca16f912.png)

iromeo:
Done 
![image](https://user-images.githubusercontent.com/72933/62644206-405ffc80-b952-11e9-8697-d5d3f26bab28.png)

PR #81, opened 2019-07-03
Issue 72

sharkovadarya:
Keyword arguments duplication https://github.com/JetBrains-Research/snakecharm/issues/72


src/main/kotlin/com/jetbrains/snakecharm/lang/validation/SnakemakeSyntaxErrorAnnotator.kt
```diff
@@ -0,0 +1,65 @@
...
+            return
+        }
+
+        val argumentNamesSet = HashSet<String>()
+        var encounteredKeywordArgument = false
+
+        st.argumentList?.arguments?.forEach {
+            // check argument name
+            val name = it.name
+            if (argumentNamesSet.contains(name)) {
```
iromeo: Could you refactor this method to make keyword argument processing more explicit? At the moment it isn't clear why we check `argumentNamesSet.contains(null)` in some cases, although it seems never can happen. Also `getName()` is method from `PsiElementBase` not clear how it is related to keyword arguments

E.g.
```
if (arg is PyKeywordArgument) {
  arg.keyword?.let { name ->
   ...
  }
  .. 
} else {
   // positional argument
  ...
}
```

iromeo: Please use `a in b` instead of `b.contains(a)`

src/main/kotlin/com/jetbrains/snakecharm/lang/validation/SnakemakeSyntaxErrorAnnotator.kt
```diff
@@ -0,0 +1,65 @@
...
+import com.jetbrains.snakecharm.lang.psi.SMKRuleParameterListStatement
+import com.jetbrains.snakecharm.lang.psi.SMKRuleRunParameter
+
+object SnakemakeSyntaxErrorAnnotator : SnakemakeAnnotator() {
+    override fun visitSMKRuleParameterListStatement(st: SMKRuleParameterListStatement) {
+        if (!SnakemakeLanguageDialect.isInsideSmkFile(st)) {
+            return
+        }
+
+        val argumentNamesSet = HashSet<String>()
```
iromeo: Let's rename to `seenKeywords`

src/main/kotlin/com/jetbrains/snakecharm/lang/validation/SnakemakeSyntaxErrorAnnotator.kt
```diff
@@ -0,0 +1,65 @@
...
+            }
+        }
+    }
+
+    override fun visitSMKRule(smkRule: SMKRule) {
+        var executionSectionOccurred = false
+
+        val sections = smkRule.getSections()
+        for (st in sections) {
+            if (st is SMKRuleParameterListStatement) {
```
iromeo: Probably `when (st)` will make code nicer

src/main/kotlin/com/jetbrains/snakecharm/lang/validation/SnakemakeSyntaxErrorAnnotator.kt
```diff
@@ -0,0 +1,65 @@
...
+object SnakemakeSyntaxErrorAnnotator : SnakemakeAnnotator() {
+    override fun visitSMKRuleParameterListStatement(st: SMKRuleParameterListStatement) {
+        if (!SnakemakeLanguageDialect.isInsideSmkFile(st)) {
+            return
+        }
+
+        val argumentNamesSet = HashSet<String>()
+        var encounteredKeywordArgument = false
+
+        st.argumentList?.arguments?.forEach {
```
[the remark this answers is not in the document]
sharkovadarya: This loop contains not only keyword arguments name check but also positional arguments after keyword arguments check, thus I have to iterate over all arguments, not just keyword arguments. Or do you propose I separate these checks into two loops: one for keyword argument names, another for positional argument after keyword argument check?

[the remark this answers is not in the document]
iromeo: Yes, I agree, I've realized that method used both keyword and positional arguments a bit later. I thought I had deleted this comment, sorry.

PR #85, opened 2019-07-04
Issue 43

sharkovadarya:
'shell' section: resolve & completion for `params` args https://github.com/JetBrains-Research/snakecharm/issues/43


src/main/kotlin/com/jetbrains/snakecharm/lang/psi/SMKParamsReference.kt
```diff
@@ -30,18 +30,18 @@ class SMKParamsReference(
...
         val variants = mutableListOf<LookupElement>()
-        getKeywordArguments()
-                ?.filter { it.name != null }
-                ?.forEach {
-                    variants.add(LookupElementBuilder.create(it.name!!).withIcon(PlatformIcons.PARAMETER_ICON))
-                }
+        variants.addAll(getKeywordArguments()
+                ?.map { arg ->
+                    LookupElementBuilder.create(arg.keyword!!).withIcon(PlatformIcons.PARAMETER_ICON)
+                } ?: emptyList())
```
iromeo: Are you sure that `arg.keyword` never returns `null`?

src/main/kotlin/com/jetbrains/snakecharm/lang/psi/SMKParamsReference.kt
```diff
@@ -29,13 +24,10 @@ class SMKParamsReference(
...
     override fun getVariants(): Array<Any> {
-        val variants = mutableListOf<LookupElement>()
-        variants.addAll(getKeywordArguments()
+        return (getKeywordArguments()
                 ?.map { arg ->
                     LookupElementBuilder.create(arg.keyword!!).withIcon(PlatformIcons.PARAMETER_ICON)
-                } ?: emptyList())
-        return variants.toTypedArray()
-
+                } ?: emptyList()).toMutableSmartList().toTypedArray()
```
iromeo: Why do we need to copy result to some intermediate `toMutableSmartList()` list ?

src/main/kotlin/com/jetbrains/snakecharm/lang/psi/SMKParamsReference.kt
```diff
@@ -29,13 +24,10 @@ class SMKParamsReference(
     }
 
     override fun getVariants(): Array<Any> {
-        val variants = mutableListOf<LookupElement>()
-        variants.addAll(getKeywordArguments()
+        return (getKeywordArguments()
```
iromeo: This could be converted to expression function

Issue #88, opened 2019-07-04
Subworflow redeclaration inspection


iromeo:
Warn if subworflow was imported with same name several times. The warning type (warning or error) and message should be consistent with runtime behaviour.

iromeo:
@nekit-000000  Please add a screenshot, close the issue and mark it as done in  #90 task

nekit-000000:
![Screenshot from 2019-07-05 10-46-35](https://user-images.githubusercontent.com/25721619/60721302-25a2fe00-9f36-11e9-83c9-a6f8f3a0f6a8.png)

Issue #83, opened 2019-07-03
Inspection: subworkflow 'workdir', 'snakefile', 'configfile' could accept only single arg


iromeo:
Ensure that it is true. Better to implement using annotator and highlight as an syntax error

iromeo:
Is related to #66

iromeo:
@nekit-000000  Please add a screenshot, close the issue and mark it as done in  #90 task

nekit-000000:
![Screenshot from 2019-07-05 15-05-37](https://user-images.githubusercontent.com/25721619/60721393-6c90f380-9f36-11e9-9705-9b027affa5d2.png)

PR #89, opened 2019-07-04
#88 Subworkflow inspections

nekit-000000:
related to #88, #83

### DIFF
--- a/src/main/kotlin/com/jetbrains/snakecharm/inspections/SmkSubworkflowRedeclarationInspection.kt
+++ b/src/main/kotlin/com/jetbrains/snakecharm/inspections/SmkSubworkflowRedeclarationInspection.kt
@@ -1,6 +1,7 @@
 package com.jetbrains.snakecharm.inspections
 
 import com.intellij.codeInspection.LocalInspectionToolSession
+import com.intellij.codeInspection.ProblemHighlightType
 import com.intellij.codeInspection.ProblemsHolder
 import com.jetbrains.snakecharm.SnakemakeBundle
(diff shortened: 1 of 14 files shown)


src/main/kotlin/com/jetbrains/snakecharm/inspections/SmkSubworkflowRedeclarationInspection.kt
```diff
@@ -16,7 +17,9 @@ class SmkSubworkflowRedeclarationInspection : SnakemakeInspection() {
         override fun visitSMKSubworkflow(subworkflow: SmkSubworkflow) {
             val nameNode = subworkflow.getNameNode()?.psi ?: return
```
iromeo: Use `subforkflow.name` instead of `subworkflow.getNameNode()?.psi.text`, better not to expose ASTNode related interface out of PsiElements. PsiElements - are containters for high level inteface for iteracting parsed elements with other code. ASTNodes - are too low-level API

iromeo: Probably better to replace `collectSubworkflows(): List<Pair<String, PsiElement>>` with `collectSubworkflows(): List<Pair<String, SmkSubworkflow>>` and do the same for rules.

iromeo: And check smth like this `subworkflow !== subworkflows.findLast { it.name = name }.second`

Issue #87, opened 2019-07-04
Params completion/resolve doesn't work for checkpoint rules


iromeo:
![image](https://user-images.githubusercontent.com/72933/60665132-4f8fee00-9e6c-11e9-8497-0c121e76969e.png)

iromeo:
![image](https://user-images.githubusercontent.com/72933/60813271-0c02f000-a19c-11e9-9bee-048cf785b421.png)

PR #91, opened 2019-07-05
Issue 87

sharkovadarya:
Params completion/resolve doesn't work for checkpoint rules https://github.com/JetBrains-Research/snakecharm/issues/87


src/main/kotlin/com/jetbrains/snakecharm/codeInsight/completion/SMKKeywordCompletionContributor.kt
```diff
@@ -50,6 +49,7 @@ object WorkflowTopLevelKeywordsProvider : CompletionProvider<CompletionParameter
     val CAPTURE = psiElement()
             .inFile(SMKKeywordCompletionContributor.IN_SNAKEMAKE)
             .andNot(SMKKeywordCompletionContributor.IN_RULE)
+            .andNot(SMKKeywordCompletionContributor.IN_CHECKPOINT)
```
iromeo: Please add a test for this

src/test/resources/com/jetbrains/snakecharm/cucumber/completion/keywords.feature
```diff
@@ -225,3 +225,84 @@ Feature: Completion for snakemake keyword-like things
...
+      | sh  | shadow               |
+      | co  | conda                |
+      | cw  | cwl                  |
+      | sc  | script               |
+      | sh  | shell                |
+      | run | run                  |
+      | wr  | wrapper              |
+
+  Scenario: Complete and replace at checkpoint level
+    Given a snakemake project
```
iromeo: Also, parametrize rule/checkpoint in other tests if its possible

src/test/resources/com/jetbrains/snakecharm/cucumber/completion/keywords.feature
```diff
@@ -254,61 +190,67 @@ Feature: Completion for snakemake keyword-like things
...
-      | run | run                  |
-      | wr  | wrapper              |
+      | rule_like  | str    | result               |
+      | rule       | inp    | input                |
+      | checkpoint | inp    | input                |
+      | rule       | out    | output               |
+      | checkpoint | out    | output               |
+      | rule       | par    | params               |
+      | checkpoint | lo     | log                  |
+      | rule       | re     | resources            |
```
iromeo: Cucumber doesn't allow to multiply params values, e.g. generate 2 (rule_like) x 23 (str & result) combinations. But here I don't see much sense in testing half features for rule and half for checkpoints. So I see two approaches 1) test full list for 'rule' and check several cases for 'checkpoint' 2) test everything for rule and checkpoint.

iromeo: I don't have any ideas how to avoid any copy paste here =/

Issue #82, opened 2019-07-03
Resolve/completion for `checkpoints` names


iromeo:
Similar to  #59 
```
# the checkpoint that shall trigger re-evaluation of the DAG
checkpoint clustering:
    input:
        "samples/{sample}.txt"
    output:
        clusters=directory("clustering/{sample}")
    shell:
        "mkdir clustering/{wildcards.sample}; "
        "for i in 1 2 3; do echo $i > clustering/{wildcards.sample}/$i.txt; done"

checkpoints.clustering.get(**wildcards).output[0]
```

iromeo:
@nekit-000000 please add a screenshot

PR #94, opened 2019-07-05
#82 Resolve and completion for 'checkpoints' names

nekit-000000:
#82


src/main/kotlin/com/jetbrains/snakecharm/lang/psi/types/SmkCheckPointsType.kt
```diff
@@ -0,0 +1,68 @@
...
+
+    override fun resolveMember(
+            name: String,
+            location: PyExpression?,
+            direction: AccessDirection,
+            resolveContext: PyResolveContext
+    ): List<RatedResolveResult> {
+
+        if (!SnakemakeLanguageDialect.isInsideSmkFile(location)) {
+            return mutableListOf()
```
iromeo: Maybe `emptyList()`? we don't expect smb will edit this list (add/remove items)

Issue #93, opened 2019-07-05
Subworkflow rename bug


nekit-000000:
On subworkflow rename
```
java.lang.ClassCastException: com.intellij.psi.impl.source.PsiPlainTextFileImpl cannot be cast to com.jetbrains.snakecharm.lang.psi.SnakemakeFile
	at com.jetbrains.snakecharm.inspections.SmkSubworkflowRedeclarationInspection$buildVisitor$1.<init>(SmkSubworkflowRedeclarationInspection.kt:16)
	at com.jetbrains.snakecharm.inspections.SmkSubworkflowRedeclarationInspection.buildVisitor(SmkSubworkflowRedeclarationInspection.kt:14)
	at com.jetbrains.snakecharm.inspections.SmkSubworkflowRedeclarationInspection.buildVisitor(SmkSubworkflowRedeclarationInspection.kt:9)
	at com.intellij.codeInspection.InspectionEngine.createVisitorAndAcceptElements(InspectionEngine.java:53)
	at com.intellij.codeInsight.daemon.impl.LocalInspectionsPass.runToolOnElements(LocalInspectionsPass.java:292)
	at com.intellij.codeInsight.daemon.impl.LocalInspectionsPass.lambda$null$5(LocalInspectionsPass.java:260)
	at com.intellij.util.AstLoadingFilter.disallowTreeLoading(AstLoadingFilter.java:126)
	at com.intellij.util.AstLoadingFilter.disallowTreeLoading(AstLoadingFilter.java:115)
	at com.intellij.codeInsight.daemon.impl.LocalInspectionsPass.lambda$visitPriorityElementsAndInit$6(LocalInspectionsPass.java:257)
	at com.intellij.concurrency.ApplierCompleter.execAndForkSubTasks(ApplierCompleter.java:133)
	at com.intellij.openapi.application.impl.ApplicationImpl.tryRunReadAction(ApplicationImpl.java:1163)
	at com.intellij.concurrency.ApplierCompleter.lambda$wrapInReadActionAndIndicator$1(ApplierCompleter.java:105)
	at com.intellij.openapi.progress.impl.CoreProgressManager.registerIndicatorAndRun(CoreProgressManager.java:582)
	at com.intellij.openapi.progress.impl.CoreProgressManager.executeProcessUnderProgress(CoreProgressManager.java:532)
[...]

Issue #60, opened 2019-06-23
Resolve/Completion for subworflows names


iromeo:
```
subworkflow otherworkflow:
    workdir:
        "../path/to/otherworkflow"
    snakefile:
        "../path/to/otherworkflow/Snakefile"
    configfile:
        "path/to/custom_configfile.yaml"

rule a:
    input:
        otherworkflow("test.txt")
    output: ...
    shell:  ...
```

iromeo:
Resolve/completion already works here (likely because subworkflow is named psi element?), test is added in https://github.com/JetBrains-Research/snakecharm/pull/95

PR #95, opened 2019-07-05
#93 Subworkflow rename bug

nekit-000000:
#93


src/test/resources/com/jetbrains/snakecharm/cucumber/completion/subworkflow_name.feature
```diff
@@ -0,0 +1,33 @@
+Feature: Completion for subworkflow name
```
iromeo: Testo for #60

Issue #56, opened 2019-05-28
Completion / Resolve for rules name in 'ruleorder' and 'localrules'


iromeo:
```
localrules: all, foo

rule all:
    input: ...

rule foo:
    ...

rule bar:
    ...

ruleorder: rule1 > rule2 > rule3
```

iromeo:
subtask of #6

iromeo:
![image](https://user-images.githubusercontent.com/72933/60995919-42876900-a35c-11e9-8b50-04921fa380e4.png)

PR #108, opened 2019-07-08
Issue 56

sharkovadarya:
Completion / Resolve for rules name in 'ruleorder' and 'localrules' https://github.com/JetBrains-Research/snakecharm/issues/56


src/main/kotlin/com/jetbrains/snakecharm/lang/psi/SMKRuleReference.kt
```diff
@@ -23,5 +22,25 @@ class SMKRuleReference(node: ASTNode): PyElementImpl(node), PsiNamedElement {
         return this
     }
 
+    override fun getReference(): PsiReference? = SMKRulePSIReference(this, TextRange(0, textLength))
+
     private fun getNameNode() = getIdentifierNode(node)
+}
+
+class SMKRulePSIReference(
```
iromeo: Psi element is called `SMKRuleReference` and reference called  `SMKRule`**PSI**`Reference` - it is very not intuitive. Sounds like `SMKRuleReference` is some reference and `SMKRulePSIReference` is some psi element.

I suggest using `SmkReferenceExpression` for psi elment (like `PyReferenceExpression`) and `SmkRuleOrCheckpointNameReference` for reference.

src/main/kotlin/com/jetbrains/snakecharm/lang/psi/SMKRuleReference.kt
```diff
@@ -23,5 +22,25 @@ class SMKRuleReference(node: ASTNode): PyElementImpl(node), PsiNamedElement {
...
+        element: PsiElement,
+        textRange: TextRange
+) : PsiReferenceBase<PsiElement>(element, textRange) {
+    private val key: String = element.text
+
+    override fun resolve(): PsiElement? =
```
iromeo: Consider implementing `PsiPolyVariantReference` for multi-resolve (e.g. if several names with same name)

src/main/kotlin/com/jetbrains/snakecharm/codeInsight/completion/SMKRuleReferenceCompletionContributor.kt
```diff
@@ -0,0 +1,37 @@
...
+import com.jetbrains.snakecharm.lang.psi.SMKWorkflowRuleOrderStatement
+import com.jetbrains.snakecharm.lang.psi.SnakemakeFile
+
+class SMKRuleReferenceCompletionContributor : CompletionContributor() {
+    init {
+        extend(CompletionType.BASIC, RuleReferenceCompletionProvider.CAPTURE, RuleReferenceCompletionProvider)
+    }
+}
+
+object RuleReferenceCompletionProvider : CompletionProvider<CompletionParameters>() {
```
iromeo: PsiReference provide `getVariants()` method. Do we really need completion contributor here?

src/test/resources/com/jetbrains/snakecharm/cucumber/resolve/localrules_and_ruleorder.feature
```diff
@@ -0,0 +1,51 @@
...
+
+    ruleorder: aaaa > <text>
+    """
+    When I put the caret after <ptn>
+    Then reference should resolve to "<symbol_name>" in "<file>"
+
+    Examples:
+      | ptn             | text      | symbol_name | file         |
+      | > bbb           | bbbb      | bbbb        | foo.smk      |
+      | > ccc           | cccc      | cccc        | foo.smk      |
```
iromeo: Please also check case `ruleorder: aa<caret>aa`. Also please check case when `ruleorder` declaration is before all rules, e.g. to be sure that our rules collector doesn't ignores rules below the declaration

src/test/resources/com/jetbrains/snakecharm/cucumber/resolve/localrules_and_ruleorder.feature
```diff
@@ -0,0 +1,51 @@
...
+    localrules: <text>
+    """
+    When I put the caret after <ptn>
+    Then reference should resolve to "<symbol_name>" in "<file>"
+
+    Examples:
+      | ptn             | text      | symbol_name | file         |
+      | localrules: aaa | aaaa      | aaaa        | foo.smk      |
+      | localrules: bbb | bbbb      | bbbb        | foo.smk      |
+      | localrules: ccc | cccc      | cccc        | foo.smk      |
```
iromeo: Please check case `localrules: aaa, b<caret>bb`. Also check that resolve works when rule declaration is in the beginning of file before other rules.

src/main/kotlin/com/jetbrains/snakecharm/codeInsight/completion/SMKRuleReferenceCompletionContributor.kt
```diff
@@ -0,0 +1,37 @@
...
+            context: ProcessingContext,
+            result: CompletionResultSet
+    ) {
+        val element = parameters.position
+        val rules = PsiTreeUtil.getParentOfType(element, SnakemakeFile::class.java)?.collectRules() ?: emptyList()
+        val checkpoints = PsiTreeUtil.getParentOfType(element, SnakemakeFile::class.java)
+                ?.collectCheckPoints()
+                ?: emptyList()
+        result.addAllElements(rules.map { it.first }.map { LookupElementBuilder.create(it) })
+        result.addAllElements(checkpoints.map { it.first }.map { LookupElementBuilder.create(it) })
```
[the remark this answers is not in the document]
sharkovadarya: If I understand correctly, it compiles without any errors:
image

src/main/kotlin/com/jetbrains/snakecharm/lang/psi/SMKRuleReference.kt
sharkovadarya: If I understand correctly, if there are multiple rules a reference can be resolved to, that means all of those rules are in different files. How should I go about collecting rules from all files in the project? I could do so in a straightforward manner, but I remember something about caching was previously mentioned.

src/main/kotlin/com/jetbrains/snakecharm/codeInsight/completion/SMKRuleReferenceCompletionContributor.kt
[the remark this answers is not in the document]
iromeo: What about runtime execution errors, not compile one, does it works correctly?

src/main/kotlin/com/jetbrains/snakecharm/lang/psi/SMKRuleReference.kt
iromeo: Or It could be several rules with the same name in the same file. Such case will be highlighted as error, but resolve could work in any case. I suggest implement it without collecting references from the same file, e.g get rules from SmkRule object.

src/main/kotlin/com/jetbrains/snakecharm/codeInsight/completion/SMKRuleReferenceCompletionContributor.kt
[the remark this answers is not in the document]
sharkovadarya: This example, executed with `snakemake -s file.smk`, runs without any errors, `output.txt` was generated, it worked.
image

src/main/kotlin/com/jetbrains/snakecharm/lang/psi/SMKRuleReference.kt
sharkovadarya: >e.g get rules from SmkRule object

I don't understand what this means. Did you mean `SnakemakeFile`? How do I get rules from `SmkRule`?

So, I have to get the project from the `SnakemakeFile` that contains the current element, go over all .smk files in that project path, and collect references from each of those files except for the current one, is that what you meant?

src/main/kotlin/com/jetbrains/snakecharm/lang/psi/SmkReferenceExpression.kt
```diff
@@ -48,6 +51,9 @@ class SmkReferenceExpression(node: ASTNode): PyElementImpl(node), PsiNamedElemen
             return variants.toTypedArray()
         }
 
+        override fun handleElementRename(newElementName: String): PsiElement =
+                element.let { (it as PsiNamedElement).setName(newElementName) }
```
iromeo: We don't need a cast here, you get "element" from constructor, so please use ther element with proper type.

src/main/kotlin/com/jetbrains/snakecharm/lang/psi/SMKRuleReference.kt
iromeo: > Did you mean SnakemakeFile

Yes

> go over all .smk files in that project path

No, just same SMK file at the moment.

src/main/kotlin/com/jetbrains/snakecharm/codeInsight/completion/SMKRuleReferenceCompletionContributor.kt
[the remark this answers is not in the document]
iromeo: ok, good.

src/main/kotlin/com/jetbrains/snakecharm/lang/psi/SmkReferenceExpression.kt
```diff
@@ -29,15 +27,22 @@ class SmkReferenceExpression(node: ASTNode): PyElementImpl(node), PsiNamedElemen
...
+    ) : PsiReferenceBase<PsiElement>(element, textRange), PsiPolyVariantReference {
         private val key: String = element.text
 
         override fun resolve(): PsiElement? =
                 getRules().firstOrNull { it.first == key }?.second ?:
                 getCheckpoints().firstOrNull { it.first == key }?.second
 
+        override fun multiResolve(incompleteCode: Boolean): Array<ResolveResult> {
+            val elements = mutableListOf<PsiElement>()
+            elements.addAll(getRules().filter { it.first == key }.map { it.second })
```
iromeo: Same as 
```
override fun multiResolve(incompleteCode: Boolean): Array<ResolveResult> =
                (getRules() + getCheckpoints())
                        .filter { (name, _) -> name == key }
                        .map { (_, psi) -> RatedResolveResult(RatedResolveResult.RATE_NORMAL, psi) }
                        .toTypedArray()
```

I prefer my version - less duplicaton and `+` already defined for lists, no need in `list.addAll()`

src/main/kotlin/com/jetbrains/snakecharm/lang/psi/SmkReferenceExpression.kt
```diff
@@ -29,15 +27,22 @@ class SmkReferenceExpression(node: ASTNode): PyElementImpl(node), PsiNamedElemen
...
 
     private class SmkRuleOrCheckpointNameReference(
-            element: PsiElement,
+            private val element: PsiNamedElement,
             textRange: TextRange
-    ) : PsiReferenceBase<PsiElement>(element, textRange) {
+    ) : PsiReferenceBase<PsiElement>(element, textRange), PsiPolyVariantReference {
         private val key: String = element.text
 
         override fun resolve(): PsiElement? =
```
iromeo: Better to implement resolve using multi resolve, e.g. `override fun resolve() = multiResolve(false).firstOrNull()?.element`

Issue #101, opened 2019-07-07
Warn if "rules." is accessed to a rule name not defined yet


iromeo:
This code on runtime fails with: 'Rules' object has no attribute 'foo'
```
rule all:
    input: rules.foo

rule foo:
    output: "out.txt"
```
![image](https://user-images.githubusercontent.com/72933/60769089-68024180-a0d4-11e9-8f34-55e03d12eb3a.png)

P.S: not sure about "rules." in `run` section, likely will be ok

iromeo:
Same issue if rule references itself - 
![image](https://user-images.githubusercontent.com/72933/60887711-b000a000-a25d-11e9-8793-f2c21f23a330.png)

iromeo:
but it works if reference same rule from `run` section

iromeo:
![image](https://user-images.githubusercontent.com/72933/60916329-beb77900-a296-11e9-9cc7-b061180aa979.png)

PR #117, opened 2019-07-09
#101 Undefined name warning

nekit-000000:
#101

### DIFF
--- a/src/main/kotlin/com/jetbrains/snakecharm/inspections/SmkYetUndefinedNameInspection.kt
+++ b/src/main/kotlin/com/jetbrains/snakecharm/inspections/SmkYetUndefinedNameInspection.kt
@@ -2,10 +2,13 @@ package com.jetbrains.snakecharm.inspections
 
 import com.intellij.codeInspection.LocalInspectionToolSession
 import com.intellij.codeInspection.ProblemsHolder
+import com.intellij.psi.util.PsiTreeUtil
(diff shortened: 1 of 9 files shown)


src/main/kotlin/com/jetbrains/snakecharm/inspections/SmkYetUndefinedNameInspection.kt
```diff
@@ -18,16 +21,21 @@ class SmkYetUndefinedNameInspection : SnakemakeInspection() {
...
             val parent = node!!.parent
             val resolvedNode = parent.reference?.resolve() // Either corresponding rule or checkpoint or null
 
-            if (resolvedNode != null && resolvedNode.textOffset > parent.textOffset) {
-                registerProblem(parent,SnakemakeBundle.message("INSP.NAME.undefined.name") +
-                                ": ${(node.parent as PyReferenceExpression).name}")
+            if (resolvedNode != null && resolvedNode.containingFile == node.containingFile &&
+                (resolvedNode.textOffset > parent.textOffset ||
+                 resolvedNode === PsiTreeUtil.getParentOfType(node, SMKRule::class.java))) {
+                val identifier = node.nextSibling.nextSibling
```
iromeo: as always it is too low-level api and not stable, use public interfaces, here is `PyReferenceExpression.nameElement`

Issue #96, opened 2019-07-05
Parsing error when new line begins with a separator in argument list


sharkovadarya:
Example:

    rule rule12:
        output: "12"
             , "13"
or

    ruleorder:
         rule4 > rule10
             >  rule7 > rule11

both of these cases result in a parsing error but snakemake compiles such code.

iromeo:
Done

PR #121, opened 2019-07-11
Issue 96

sharkovadarya:
Parsing error when new line begins with a separator in argument list https://github.com/JetBrains-Research/snakecharm/issues/96


src/main/kotlin/com/jetbrains/snakecharm/lang/parser/SnakemakeExpressionParsing.kt
```diff
@@ -65,17 +66,52 @@ class SnakemakeExpressionParsing(context: SnakemakeParserContext) : ExpressionPa
...
+                 * e.g. atToken(), matchToken(), etc.,
+                 * it might apply filter and change the token type,
+                 * while nextToken() does not do that,
+                 * it just increases the current lexeme number, doesn't look at the token
+                 */
+                if (indents == 0 &&
+                        !findTokenRawLookup(separatorTokenType) &&
+                        !findTokenRawLookup(PyTokenTypes.INDENT) &&
+                        !findTokenRawLookup(PyTokenTypes.INCONSISTENT_DEDENT) &&
+                        !findTokenRawLookup(PyTokenTypes.DEDENT)) {
```
iromeo: Seems here we do 4 almost same lookups (skipping all whitespaces each time) instead of 1 lookup with checking that element not from token set. 

Could we implement it like `Parsing.atAnyOfTokens()` or use token set here for 4 elements ?
I'd rather expect smth like
*  `findTokenRawLookup(separatorTokenType, PyTokenTypes.INDENT, PyTokenTypes.INCONSISTENT_DEDENT, PyTokenTypes.DEDENT)` 
or
* `findTokenRawLookup(some_reasonable_named_token_set)`

Issue #102, opened 2019-07-07
Params completion for shell cmd in run section


iromeo:
Params completion for shell cmd in run section
```
rule foo:
   params: a=1
   run:
      shell("echo {params.a}")
```
Here completion expected same as in 'shell' section

PR #123, opened 2019-07-11
Issue 102

sharkovadarya:
Params completion for shell cmd in run section https://github.com/JetBrains-Research/snakecharm/issues/102


src/main/kotlin/com/jetbrains/snakecharm/codeInsight/SnakemakeParamsInShellReferenceContributor.kt
```diff
@@ -5,20 +5,30 @@ import com.intellij.patterns.PlatformPatterns
...
 import java.util.regex.Pattern
 
 class SnakemakeParamsInShellReferenceContributor : PsiReferenceContributor() {
+    private val insideRuleSection = PlatformPatterns.psiElement()
+            .inFile(SMKKeywordCompletionContributor.IN_SNAKEMAKE)
+            .inside(SMKRuleParameterListStatement::class.java)
+    private val insideCallExpressionInRuleRunParameter = PlatformPatterns.psiElement()
+            .inFile(SMKKeywordCompletionContributor.IN_SNAKEMAKE)
+            .inside(PyCallExpression::class.java)
+            .withAncestor(6, PlatformPatterns.psiElement(SMKRuleRunParameter::class.java))
```
iromeo: we could easily break such low-level checks, e.g. place "shell" call inside one of several `if` statements. 

NB: Is always please do not use low level impl details of PSI / AST tree outside of PsiElements. PSI and AST could be changed in future or by language injections or etc, it shouldn't break all features at once due to hardcoded things like this.

src/main/kotlin/com/jetbrains/snakecharm/codeInsight/SnakemakeParamsInShellReferenceContributor.kt
```diff
@@ -30,13 +40,23 @@ class SnakemakeParamsInShellReferenceContributor : PsiReferenceContributor() {
                         val paramsMatcher = paramsPattern.matcher(element.text)
 
                         val isShellCommand = PsiTreeUtil
-                                .getParentOfType(element, SMKRuleParameterListStatement::class.java)!!
-                                .section?.textMatches(SnakemakeNames.SECTION_SHELL) == true
+                                .getParentOfType(element, SMKRuleParameterListStatement::class.java)
```
iromeo: now is ok to use kotlin ext for `parent of type`

Issue #57, opened 2019-06-21
Inspection: wildcards in shell script can be used only with `wildcards.` prefix.


olegs:
Snakemake error message is quite misleading, 

This won't work with strange error `sample` is unknown. 
```
rule trim_single_fastq:
    input: os.path.join(config['fastq_dir'], '{sample}.fastq')
    output: 'cleaned/{sample}_se.fastq'
    threads: 4
    conda: 'envs/bio.env.yaml'
    shell: 'trim_galore --cores {threads} --nextera {input} -o cleaned/ &> {log}; ' \
           'mv cleaned/{sample}_val.fq {output}'
```

Changing `cleaned/{sample}_val.fq` to `cleaned/{wildcards.sample}_val.fq` fixes the problem.

iromeo:
Done

![image](https://user-images.githubusercontent.com/72933/64910627-16ec7a80-d721-11e9-80df-429b642a220d.png)

![image](https://user-images.githubusercontent.com/72933/64910634-28ce1d80-d721-11e9-9bd8-65c0ecc9e488.png)

Issue #53, opened 2019-05-24
Support highlighting for arguments of format strings


iromeo:
Snakemake automatically substitutes input, output, wildcards, etc. objects into strings used in input/output/message/shell, etc. sections. So it is a bit similar to f-strings appeared in python 3.6. At the moment no highlighting is substitution syntax is provided:
 
![image](https://user-images.githubusercontent.com/72933/58317807-3fbdcc80-7e1f-11e9-8bb5-701202f45f94.png)

Let's support highlighting similar to that which is provided by PyCharm for python f-strings, e.g.:

![image](https://user-images.githubusercontent.com/72933/58314553-688e9380-7e18-11e9-894c-ed2f3188c998.png)

P.S: we cannot use f-strings instead of ordinary strings in snakemake due to undefined variable errors, some vars appear lately on runtime

iromeo:
Features like #43 could be easily implemented if we will detect substitution at parser level (like f-string support in PyCharm)

iromeo:
Format strings use Python Mini Language Format, see  https://docs.python.org/3/library/string.html

Let's try to implement this using language injection approach.
* here format language is much more simple than syntax of f-strings
* mini language format isn't a valid python, e.g. in `{foo.some_dict[key]}` quotes around key are omitted
* we'd like to support this not for all strings in *.smk file, but only for some cases, e.g. args of sections, "shell" call, etc.
* we don't need to many features there, e.g. highlighting, completion, resolve, inline/extract param are enough
* completion/resolve isn't very complicated there, not required to make python completion work there

iromeo:
Some examples:

Wildcards regexp constraints:
```
rule index_bams:
    input: '{anywhere}/{sample}.bam'
    output: '{anywhere}/{sample, [^/]*}.bam.bai'
    wrapper: '0.31.1/bio/samtools/index'

```

Escaped `{` and `}`:
```
rule bam_qc_pbc_nrf:
    input: rules.bam_to_pileup.output
    output: 'qc/pbc_nrf/{sample}.pbc_nrf.tsv'
    params:
          tmp_dir='tmp'
    shell: '''
mkdir -p {params.tmp_dir} &&
(T=$'\\t'
>&2 echo "TotalReadPairs${{T}}DistinctReadPairs${{T}}OneReadPair${{T}}TwoReadPairs${{T}}\
[...]

PR #135, opened 2019-07-22
Issue 57

sharkovadarya:
Inspection: wildcards in shell script can be used only with `wildcards.` prefix https://github.com/JetBrains-Research/snakecharm/issues/57


iromeo:
I think we need to re-implement the logic of the inspection after @nekit-000000 pull request #53 has been accepted. Current impl highlights all unresolved identifiers instead of wildcards and doesn't cover all usecases (e.g. variables defined somewhere in snakemake sources, e.g. `config`). 

I think, that we could detect rule's wildcards set from input, output, log (and maybe smth other suitable sections) sections in a similar way as snakemake does. So then we could use this inspections for 100% wildcards.

At the moment we get false positives in files which included in some other file containing global vars + for snakemake global vars. E.g.
image

iromeo:
Depends #184

iromeo:
Feature is implemented.

Issue #16, opened 2019-01-11
Multiline strings declaration support


iromeo:
E.g. see `shell` arg in:

```
rule download_phantompeakqualtools:
    output: directory("bin/phantompeakqualtools")
    params:
        targz="phantompeakqualtools.tar.gz"
    shell: 'cd bin && '
           'curl --location '
           'https://storage.googleapis.com/google-code-archive-downloads/v2/'
           'code.google.com/phantompeakqualtools/ccQualityControl.v.1.1.tar.gz '
           '--output {params.targz} && '
           'tar xvf {params.targz}'
```

iromeo:
Workaround, use `\` line continuation:

```
   shell: "bo --config bo_config_dir download -w " \
          "Clean/A01/V300010918_L4_HUMvotMAAAAAAA-14_*.fq.gz fastq"
```

PR #136, opened 2019-07-23
Issue 16

sharkovadarya:
Multiline strings declaration support https://github.com/JetBrains-Research/snakecharm/compare/16


src/main/kotlin/com/jetbrains/snakecharm/lang/parser/SnakemakeExpressionParsing.kt
```diff
@@ -158,19 +141,7 @@ class SnakemakeExpressionParsing(context: SnakemakeParserContext) : ExpressionPa
...
-                                    nextToken()
-                                }
-                                if (atToken(PyTokenTypes.DEDENT)) {
-                                    indents--
-                                    nextToken()
-                                } else {
-                                    break
-                                }
-                            }
+                            skipDedents(1) { incorrectUnindentMarker = myBuilder.mark() }
```
iromeo: Theoretically ` incorrectUnindentMarker = myBuilder.mark()` could be called several times and lead to unbalanced markers error, better to refactor this code to grant that we don't introduce possible unbalanced markers error in case of any unexpected PSI or changes in python code. Here possible sources of such errors are:
1) `myBuilder.mark()`  called several times but we keep reference only on one result
2) some other marker declared before `incorrectUnindentMarker` could be closed while `incorrectUnindentMarker` is still opened

src/test/kotlin/com/jetbrains/snakecharm/lang/parser/SnakemakeParsingTest.kt
```diff
@@ -221,6 +221,7 @@ class SnakemakeParsingTest : ParsingTestCase(
         doTest()
     }
 
+
```
iromeo: Please review diff before commit, this empty line was occasionally introduced in code and committed with other changes

src/main/kotlin/com/jetbrains/snakecharm/lang/parser/SnakemakeLexer.kt
```diff
@@ -31,5 +31,5 @@ class SnakemakeLexer : PythonIndentingLexer() {
         val KEYWORDS_2_TEXT = KEYWORDS.map { it.value to it.key }.toMap()
     }
 
-    override fun getTokenType() = KEYWORDS[tokenText] ?: super.getTokenType()
+    // override fun getTokenType() = KEYWORDS[tokenText] ?: super.getTokenType()
```
iromeo: This commit is completely not related to issue #16 and pull request. Let's do not mix different complicated changes in one pull request

src/main/kotlin/com/jetbrains/snakecharm/lang/parser/SnakemakeStatementParsing.kt
```diff
@@ -29,6 +30,39 @@ class SnakemakeStatementParsing(
         futureFlag: FUTURE?
 ) : StatementParsing(context, futureFlag) {
 
+    private val keywordToParsingFunction = ImmutableMap.Builder<String, (IElementType?) -> Unit>()
```
iromeo: Why do we need this change if you remap identifier token to snakemake specific token type? Could we leave source code after remap as it was before (based on token types comparison, not on matching identifier text content)?

src/main/kotlin/com/jetbrains/snakecharm/lang/parser/SnakemakeTokenTypes.kt
```diff
@@ -81,18 +82,19 @@ object SnakemakeTokenTypes {
     )
 
     val stringToTokenType = mapOf<String, IElementType>(
```
iromeo: Is it a copy paste from SnakemakeLexer.KEYWORDS mapping?

src/main/kotlin/com/jetbrains/snakecharm/lang/parser/SnakemakeStatementParsing.kt
```diff
@@ -80,9 +114,17 @@ class SnakemakeStatementParsing(
...
+            // XXX: workaround/fix for #130:
+            val actualToken = SnakemakeTokenTypes.stringToTokenType[myBuilder.tokenText]
+            if (actualToken != null) {
+                myBuilder.remapCurrentToken(actualToken)
+            }
+        }
         val tt = myBuilder.tokenType
 
-        if (tt !in SnakemakeTokenTypes.WORKFLOW_TOPLEVEL_DECORATORS || scope.inParamArgsList) {
+        val parsingFunction = keywordToParsingFunction[myBuilder.tokenText]
```
iromeo: No clear why do we need here to check `myBuilder.tokenText` again, cacl it's hash, seems `tt = myBuilder.tokenType` already defines how to parse next piece of code.

src/main/kotlin/com/jetbrains/snakecharm/lang/parser/SnakemakeStatementParsing.kt
```diff
@@ -80,9 +114,17 @@ class SnakemakeStatementParsing(
         val scope = context.scope
 
         myBuilder.setDebugMode(false)
+        if (myBuilder.tokenType == PyTokenTypes.IDENTIFIER) {
+            // XXX: workaround/fix for #130:
+            val actualToken = SnakemakeTokenTypes.stringToTokenType[myBuilder.tokenText]
+            if (actualToken != null) {
+                myBuilder.remapCurrentToken(actualToken)
```
iromeo: What about filter method? I think it also should be updated somehow or even removed.

testData/psi/KeywordLikeIdentifiersAsIdentifiers.smk
```diff
@@ -8,4 +8,5 @@ rule foo: run:
     checkpoint = 1
     configfile = 2
     report = 3
-    wildcard_constraints = 4
\ No newline at end of file
+    wildcard_constraints = 4
+    print(rule)
```
iromeo: Could you add also a test when `print(rule)` is not only in `run` section? is `print(rule)` is example when parsing is broken due to statement break inserted by lexer? If not let's add such test. If yes, please also add testcase to lexer test because it is also a lexer-related issue.

testData/psi/RuleParamsListArgsStringMultiline.smk
```diff
@@ -1,4 +1,4 @@
-print(
+@workflow.input(
```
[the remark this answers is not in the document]
iromeo: fixed

src/main/kotlin/com/jetbrains/snakecharm/lang/parser/SnakemakeExpressionParsing.kt
iromeo: @sharkovadarya still actual

src/main/kotlin/com/jetbrains/snakecharm/lang/parser/SnakemakeExpressionParsing.kt
```diff
@@ -390,6 +394,9 @@ class SnakemakeExpressionParsing(context: SnakemakeParserContext) : ExpressionPa
         } else if (!atAnyOfTokensSafe(PyTokenTypes.PLUS)) {
             statementBreakMarker?.rollbackTo()
         } else {
+            if (incorrectUnindentMarker != null) {
+                incorrectUnindentMarker?.error(SnakemakeBundle.message("PARSE.incorrect.unindent"))
+            }
```
iromeo: Please simplify this `if` in 397...399. Here you do double-check for nullable value.

Issue #144, opened 2019-07-24
File reference support for `conda` section


iromeo:
Expected completion/resolve for *.yaml / *.yml file relative paths. A quick fix, which creates the missing file path is also desired

E.g.:

```
rule filter_sort_bam_paired:
    input: '{anywhere}/{sample}.bam2.raw'
    output: '{anywhere}/{sample}.bam'
    conda: 'envs/bio.env.yaml'
    resources:
        threads = 2,
        mem = 8, mem_ram = 4,
        time = 60 * 120
    shell: 'samtools view -bh -f2 -q30 {input} > {output}.filtered; samtools sort {output}.filtered -o {output}; rm {output}.filtered'

```
Depends on #20

PR #158, opened 2019-07-28
#144 File reference support for conda section

nekit-000000:
#144


nekit-000000:
Last commit is related to #159

Issue #146, opened 2019-07-25
Multiline string syntax error not reported


iromeo:
```
rule boo:
    shell: "echo hello"
    "world"
```
snake make fails with `SyntaxError in line 4 of /Users/romeo/IdeaProjects/untitled/pymodule37/foo.smk:
unexpected EOF while parsing
`
```
$ snakemake -s foo.smk --print-compilation
@workflow.rule(name='boo', lineno=1, snakefile='/Users/romeo/IdeaProjects/untitled/pymodule37/foo.smk')
@workflow.shellcmd ( "echo hello"

)
@workflow.run
def __rule_boo(input, output, params, wildcards, threads, resources, log, version, rule, conda_env, singularity_img, singularity_args, use_singularity, bench_record, jobid, is_shell, bench_iteration, shadow_dir):
        shell ( "echo hello" , bench_record=bench_record, bench_iteration=bench_iteration
) 

@workflow.docstring("world")
```

iromeo:
See #16

PR #160, opened 2019-07-30
Issue 146

sharkovadarya:
Multiline string syntax error not reported https://github.com/JetBrains-Research/snakecharm/issues/146


iromeo:
The error is the same as in snakemake, but sounds very unclear. Could we use something more readable?

Issue #147, opened 2019-07-25
False Error: `wildcards` has to be first lambda arg


iromeo:
Related to #27

```
rule boo:
    output: touch("foo.out")
    params: x = lambda a, output: f"{type(a)}, {type(output)}"
    shell:
        "echo '{params.x}'"
```

SnakeCharm shows an error, but pipleine works ok

![image](https://user-images.githubusercontent.com/72933/61868865-00753000-aee3-11e9-9fed-5f41cc1d71a4.png)

iromeo:
![image](https://user-images.githubusercontent.com/72933/62943021-2efb7200-bde2-11e9-90ce-28d9dbf529d0.png)

PR #163, opened 2019-07-31
Issue 147

sharkovadarya:
False Error: `wildcards` has to be first lambda arg https://github.com/JetBrains-Research/snakecharm/issues/147

Please note that only the _first_ lambda parameter technically can have any name except for `output`, `input`, `threads` or `resources`.

    params: x=lambda output: "text"

results in a `TypeError: <lambda>() got multiple values for argument '...'`; `wildcards`.

`wildcards` cannot be used as any parameter but the first.

    params: lambda a, wildcards: "text"

results in a `TypeError: <lambda>() missing 1 required positional argument: 'wildcards'`.


src/main/kotlin/com/jetbrains/snakecharm/inspections/SmkLambdaRuleParamsInspection.kt
```diff
@@ -96,7 +97,13 @@ class SmkLambdaRuleParamsInspection : SnakemakeInspection() {
                                             PyRenameArgumentQuickFix()
                                     )
                                 }
-                                return
+                            }
+                            println("$index ${pyParameter.name}")
```
iromeo: Please remove this debug output.

src/main/kotlin/com/jetbrains/snakecharm/inspections/SmkLambdaRuleParamsInspection.kt
```diff
@@ -43,14 +45,28 @@ class SmkLambdaRuleParamsInspection : SnakemakeInspection() {
...
+                                    registerProblem(
+                                            pyParameter,
+                                            SnakemakeBundle.message(
+                                                    "INSP.NAME.only.these.parameters.in.section",
+                                                    WILDCARDS_LAMBDA_PARAMETER,
+                                                    SnakemakeNames.SECTION_INPUT
+                                            ),
+                                            ProblemHighlightType.WARNING,
+                                            null,
+                                            PyRenameArgumentQuickFix()
```
iromeo: Doesn't work properly, doesn't rename usages in the right part of lambda. Although standard rename refactoring for the same element also renames all usages.

iromeo: E.g. `outdir=lambda wd, output:  wd.foo`

src/main/kotlin/com/jetbrains/snakecharm/inspections/SmkLambdaRuleParamsInspection.kt
```diff
@@ -43,14 +45,28 @@ class SmkLambdaRuleParamsInspection : SnakemakeInspection() {
...
-                                )
+                                if (index == 0) {
+                                    registerProblem(
+                                            pyParameter,
+                                            SnakemakeBundle.message(
+                                                    "INSP.NAME.only.these.parameters.in.section",
+                                                    WILDCARDS_LAMBDA_PARAMETER,
+                                                    SnakemakeNames.SECTION_INPUT
+                                            ),
+                                            ProblemHighlightType.WARNING,
```
iromeo: Let's use weak warning or maybe remove it at all. A user could use `wd` just to make the code shorter. It is ok that we suggest `wildcards` in the completion list, but there is no warning errors if user prefers shorter name for the variable. If you decide to show weak warning please explain that according to conventions it is better to use `wildcards` name for this variable. Warning `wildcards` has to be ... - isn't correct here, nobody forbids use any name here.

Issue #155, opened 2019-07-26
ruleorder, localrules should resolve/complete not only rules from same file, but all rules from project


iromeo:
E.g.:

```
# Snakefile
ruleorder rule1 > rule2

include: 1.smk
include: 2.smk
```

```
# 1.smk
rule rule1:
   input: ""
```
```
# 2.smk
rule rule2:
   input: ""
```

At the moment Snakemake highlightes rules in `Snakefile` as unresolved.

iromeo:
First completion:
![image](https://user-images.githubusercontent.com/72933/63272936-a8450a00-c2a5-11e9-8c64-a37c286e7de0.png)

2nd completion:
![image](https://user-images.githubusercontent.com/72933/63272968-b98e1680-c2a5-11e9-987f-0e8c7c326625.png)

PR #164, opened 2019-08-01
Issues 155, 173

sharkovadarya:
ruleorder, localrules should resolve/complete not only rules from same file, but all rules from project https://github.com/JetBrains-Research/snakecharm/issues/155

I assumed resolve/completion should work not for all files in the project, but only for those included in the current file via `include` in this file or other files included in this file. 

show all names only on 2nd auto-completion invocaton #173


src/main/kotlin/com/jetbrains/snakecharm/lang/psi/impl/SmkReferenceExpressionImpl.kt
```diff
@@ -60,10 +61,28 @@ class SmkReferenceExpressionImpl(node: ASTNode): PyElementImpl(node), SmkReferen
...
-                    AbstractSmkRuleOrCheckpointType.createRuleLikeLookupItem(name, elem as SmkRuleOrCheckpoint)
-                }.toTypedArray()
+        override fun getVariants(): Array<Any> {
+            val results = mutableListOf<Pair<String, SmkRuleOrCheckpoint>>()
+
+            val module = ModuleUtilCore.findModuleForPsiElement(element)
+            if (module != null) {
+                val ruleResults = mutableListOf<SmkRule>()
+                val checkpointResults = mutableListOf<SmkCheckPoint>()
+                AbstractSmkRuleOrCheckpointType.addVariantFromIndex(SmkRuleNameIndex.KEY, module, ruleResults, SmkRule::class.java)
```
iromeo: Completion and resolve should be consistent. Or resolve should at least cover items from completion. Here you use indexes for completion and includes for resolve, so you could get `unresolved` error for an item from the completion list. See AbstractSmkRuleOrCheckpointType how to implement resolve.

sharkovadarya: Sorry, I don't quite get what you mean. `getVariants()` for `SmkReferenceExpressionImpl` also uses includes, see code from line 90 up to the end of the method (applying the latest changes in which I cleaned up the code a bit). Could you please provide an example? From your message I assumed something like this:

`foo.smk`

    rule foo:
        input: ...

`boo.smk`

    rule boo: input: "..."

`file.smk`

    rule NAME: ...

    include: "foo.smk"
    localrules: NAME, <caret>

If we call completion at `<caret>`, the list will  include `foo` but not `boo` because `boo`, although indexed, is not inlcuded in `file.smk`. So this example does not result in an `unresolved reference` error that you suggest might occur here; could you please provide me with another example?

iromeo: > also uses includes, see code from line 90

Ok, I was wrong. Resolve and Completion impl are consistent, both uses index + filtration by includes.

So 2 moments:
1. The behavior doesn't match to resolve and completion after `rules.`, because it doesn't filter by includes. 
2. False unresolved errors in case:
```
# ./Snakemeke
include 'foo.smk'
include 'boo.smk'
```
```
# ./foo.smk
rule foo:
   output: "in.txt"
```

```
# ./boo.smk
ruleorder: boo > foo
rule boo:
   output: "in.txt"
```

iromeo: I think the desired behaviour for user could be like in  #173 
1. resolve works using index
2. completion is implemented using 2 modes. Default completion shows rules from this & included files (as here). Second completion shows all rules from index
3. Completion / resolve after `rules.` and for `ruleorder`, `localrules` are consistent
4. We have some smart inspection which could warn about rules which could be resolved using index but not accessible via includes (e.g. in the above example, 'foo' rule in boo.smk is accessible by includes, because both files are included into Snakefile)

sharkovadarya: >Completion / resolve after rules. and for ruleorder, localrules are consistent

Just to clarify: does this mean that resolve for `rules.` should stay as it is (as in, IIRC, all indexed rules are included) and completion for `rules.` should be modified to use includes on first invocation and all indexed rules on second invocation?

>We have some smart inspection which could warn about rules 

Does 'warn' mean inspection severity level is set to `WARNING`?

iromeo: > Just to clarify:
Yes, but it is #173. Here impl is expected to be consistent with existing features, i.e similar to `rules.` resolve/completion.

> Does 'warn' mean
It depends on how accurate our implementation is, i.e. probability of FP result. If our impl isn't very accurate or we know how to cheat it or some existing projects where we show FP, then a weak warning is better.

sharkovadarya: Regarding the aforementioned inspection: what if rule `foo` is present both in `file1.smk` and `file2.smk` and we use it in `ruleorder` in `file1.smk` without `file2.smk` being included? I take it that in this case we don't need to highlight this rule? If it wasn't present in `file1.smk`, we would need to highlight it, of course, but if it is, then does the other reference matter?

iromeo: > I take it that in this case we don't need to highlight
agree

> then does the other reference matter?
If both *.smk are included in some other file (e.g. Snakemake) we could show weak warning that this `foo` rules override each other.

if one of them is included in the other, then we could show warning/error about rule redeclaration

If files are absolutely not related, then we don't care that there is some other file in the project with a similar rule name.

P.S: If your config doesn't result in runtime error/bug, then we don't need warn user about smth.

sharkovadarya: >If both *.smk are included in some other file (e.g. Snakemake) we could show weak warning that this foo rules override each other.
IIRC this logic (determining which files the current one in included in) was a bit complicated and required smth like caching or indexing; should this be implemented here or can it be added to the inspection later?

>if one of them is included in the other, then we could show warning/error about rule redeclaration
This reminds me that we need to update `SmkRuleRedeclarationInspection` to also process rules from included files. Is it ok to show error both here and in this inspection (where the element that gets highlighted is the actual rule name) or should we leave error highlighting in `SmkRuleRedeclarationInspection` and show a warning here?

iromeo: > IIRC this logic (determining which files the current one in included in) was a bit complicated and required smth like caching or indexing; should this be implemented here or can it be added to the inspection later?

> This reminds me that we need to update SmkRuleRedeclarationInspection

I meant that it should be a separate issue, #195

Issue #152, opened 2019-07-26
'shell' section & call: resolve & completion for input args


iromeo:
Same as #43, but for `input` section

iromeo:
Obsolete, see in #183, #184, #207

Issue #3, opened 2019-01-11
Resolve/Completion for: input, output, params rule options in shell argument


iromeo:
Code completion and resolve for input, output, params rule options in shell argument
```
rule all:
    input: "multiqc_report.html"
    output: "boo"
    shell: 'echo {input} {output}'
```

iromeo:
Also: `threads`, `params`

iromeo:
Is related to #53

iromeo:
Mostly covered by #183.
Need to add Snakemake string format language injection to `shell()` command in `run:` section

iromeo:
Injection for shell added in 64b6fd3324775e33fe17598b138d2267fc7befda

PR #165, opened 2019-08-01
Issue 152

sharkovadarya:
'shell' section & call: resolve & completion for input args https://github.com/JetBrains-Research/snakecharm/issues/152

Also adds keyword arguments resolve/completion for `output`, `resources`, and `log`, which seem to be the only sections (in addition to `input` and `params`) which can have keyword arguments and be used inside a `shell` section or a call to `shell` function. So it is a partial fix for https://github.com/JetBrains-Research/snakecharm/issues/3


iromeo:
AFAIU https://github.com/JetBrains-Research/snakecharm/pull/166 includes this commits, do we need this pull request? Basically feedback for this pull request is subset of my feedback for #166

iromeo:
Is included in #166

PR #170, opened 2019-08-06
#53 support highlighting for arguments of format strings

nekit-000000:
#53


iromeo:
Injection targets:
* Is it ok to use SmkSL for wrapper, conda sections? (is enabled at the moment, but I'm not sure that it should be enabled here)
* Seems we shouldn't use it for `wildcards_constraints` section, because it defines wildcards regexps
* Injection doesn't work in args of standard snakemake utility methods, e.g.
```
rule foo
   input: 
       d = directory("{sample}")
       c = ancient("smth/{sample}.txt")
       k = expand("foo{{sample}}{group}.fq"), group = ["_1", "_2"])
   output: temp("bams/{sample}.bam.raw")
    run:
        shell("foo {sample}")
```
and other supported methods

src/main/java/com/jetbrains/snakecharm/string_language/lang/parser/_SmkSLLexer.java
```diff
@@ -0,0 +1,552 @@
+/* The following code was generated by JFlex 1.7.0 tweaked for IntelliJ platform */
```
iromeo: Please 
1. remove generated class from git repo
2. add it to git ignore file
3. updated `build.gradle` to include code generation step before compilation. I think `https://github.com/thomaslee/gradle-jflex-plugin` will help

src/main/kotlin/com/jetbrains/snakecharm/string_language/SmkStringLanguage.kt
```diff
@@ -0,0 +1,6 @@
+package com.jetbrains.snakecharm.string_language
+
+import com.intellij.lang.Language
+import com.jetbrains.snakecharm.lang.SnakemakeLanguageDialect
+
+object SmkStringLanguage : Language(SnakemakeLanguageDialect,"SmkStringLanguage")
```
iromeo: Is it really required to make SmkSL language extent Snakemake?

iromeo: Maybe `SmkSL` here for consistency with SmkSlFileType?

src/main/kotlin/com/jetbrains/snakecharm/string_language/SmkSLFileType.kt
```diff
@@ -0,0 +1,14 @@
...
+import javax.swing.Icon
+
+object SmkSLFileType : LanguageFileType(SmkStringLanguage) {
+    override fun getIcon(): Icon? = null
+
+    override fun getName() = "SmkSL"
+
+    override fun getDefaultExtension() = "smkStringLanguage"
+
+    override fun getDescription() = "Snakemake string language file"
```
iromeo: The description is shown in injection dialog and isn't very clear, lets use "Snakemake Formatted String"

image

src/main/java/com/jetbrains/snakecharm/string_language/lang/parser/_SmkSLLexer.java
nekit-000000: SmkSL lexer is built with a tweaked version of JFlex to make it compatible with IDEA. So it is easier to leave the generated class as is, furthermore it seems to be a normal practice to keep generated files in repo for other projects.

src/main/kotlin/com/jetbrains/snakecharm/string_language/SmkSLTokenTypes.kt
```diff
@@ -1,10 +1,38 @@
...
+import com.intellij.lang.*
+import com.intellij.psi.PsiElement
 import com.intellij.psi.tree.IElementType
+import com.intellij.psi.tree.ILazyParseableElementType
+import org.intellij.lang.regexp.RegExpLanguage
+import org.intellij.lang.regexp.RegExpParserDefinition
+
+class SmkSLTokenType(debugName: String) : IElementType(debugName, SmkSL)
+
+class RegExpElementType(debugName: String) : ILazyParseableElementType(debugName, SmkSL) {
```
iromeo: Let's rename to `SmkSLRegExpElementType`

src/main/kotlin/com/jetbrains/snakecharm/string_language/SmkSLTokenTypes.kt
```diff
@@ -1,10 +1,38 @@
...
+    override fun parseLight(chameleon: ASTNode): PsiBuilder {
+        val project = chameleon.psi.project
+        val regExpLanguage = RegExpLanguage.INSTANCE
+        val definition =
+                LanguageParserDefinitions.INSTANCE.forLanguage(regExpLanguage) as RegExpParserDefinition
+        val capabilities =
+                definition.defaultCapabilities
+        val lexer = definition.createLexer(project, capabilities)
```
iromeo: Not sure that default capabilities are ok for us, e.g. compare with PythonRegexpLanguage capabilities set + check runtime behavior.

Issue #171, opened 2019-08-07
Plugin indexes files from .snakemake directory


sharkovadarya:
`snakemake` script run creates a `.snakemake` directory in the working directory. There can be a `shadow` directory in `.snakemake`, which contains symlinks to files in the current workdir; this results in SnakeCharm repeatedly indexing those files and putting what may seem as multiple copies of the same rule in the index, and these copies can pop up in the completion list, confusing the user.

This can be fixed by user marking their `.snakemake` directory as excluded.

PR #172, opened 2019-08-07
Issue 171 partial workaround

sharkovadarya:
Plugin indexes files from .snakemake directory https://github.com/JetBrains-Research/snakecharm/issues/171

Plugin still indexes files from .snakemake directory, but now when the user calls autocompletion on `rules.`, there is a relative file path next to each rule in the completion list.


iromeo:
Please show path relative to rule file module/project content root, not relatively to some other rule file

Looks not clear:
image

I expect smth like:
```
rule_a  [module1]/rules/boo.smk
rule_b  [moudle1/.snakemake/shadow/tmpxxx/boo.smk
rule_c  [module1]/Snakefile
```

src/main/kotlin/com/jetbrains/snakecharm/lang/psi/types/AbstractSmkRuleOrCheckpointType.kt
```diff
@@ -133,13 +134,11 @@ abstract class AbstractSmkRuleOrCheckpointType<T: SmkRuleOrCheckpoint>(
...
     companion object {
-        fun <T: SmkRuleOrCheckpoint> createRuleLikeLookupItem(name: String, elem: T): LookupElement {
+        fun <T: SmkRuleOrCheckpoint> createRuleLikeLookupItem(name: String, elem: T, currentFile: PsiFile): LookupElement {
             val containingFileName = elem.containingFile.name
             val displayPath = try {
-                val elementFilePath = elem.containingFile.originalFile.virtualFile?.presentableUrl ?: containingFileName
-                val elementFile = File(elementFilePath)
-                val module = ModuleUtilCore.findModuleForPsiElement(elem.originalElement)
-                val currentDirectory = File(module?.moduleFilePath ?: elementFilePath).parentFile
+                val elementFile = File(elem.containingFile.originalFile.virtualFile?.presentableUrl ?: containingFileName)
```
iromeo: I think `originalFile` not needed here + please keep lines not very long

sharkovadarya: If I don't use `originalFile`, a call to `virtualFile` here returns null in completion tests in `localrules_and_ruleorder.feature` since `getViewProvider().isEventSystemEnabled()` from `PsiFileImpl` returns false. I have spent some time debugging this while working on another issue but I haven't been able to figure out why this happens (somehow only for completion, not only in tests but also in an IDEA debug instance; resolve, however, works just fine). I discovered that a call to `originalFile` solves this particular problem without having to modify anything else so I went with it. I can try to spend more time on debug but I'm not sure if it's that necessary? I assumed a call to `originalFile` is not that taxing so it's ok to use, but maybe this is not the case, and it if, of course, unclear if this really does solve the problem in every possible case.

sharkovadarya: Update: I think it happens because `containingFile` created during completion is a copy, not the original file, and `isEventSystemEnabled()` should return false for that copy (see `CompletionInitializationUtil`, `insertDummyIdentifier()` method, line 123: the original file is copied to the `hostCopy` variable in `obtainFileCopy()` method, and in this method there is a check on line 183 that `isEventSystemEnabled()` should return false and this is intended). However, since we need access to the actual virtual file, we have to use `originalFile` and not this copy we're getting, I suppose. The file created as `hostCopy` and the file we get by calling `elem.containingFile` are the same object, I checked in debug.

iromeo: Ok, thanks for the explanation, so it is needed.

sharkovadarya:
@iromeo in case the 'waiting for reply' label refers to incorrect choice of folder to relativize file path against: I tried to fix this in the latest commit. If it refers to something else, then could you please tell me what? I can't find other change requests.

iromeo:
> @iromeo in case the 'waiting for reply' label refers to incorrect choice of folder to relativize file path against: I tried to fix this in the latest commit. If it refers to something else, then could you please tell me what? I can't find other change requests.

Sorry, it was my fault because I've forgot to push "submit review" button. I've added the label to mark issues where I've pushed some feedback to differ them from pull requests without feedback or where it is processed. Fill free to remove the label after you've handled my feedback.

src/main/kotlin/com/jetbrains/snakecharm/lang/psi/types/AbstractSmkRuleOrCheckpointType.kt
```diff
@@ -143,18 +144,17 @@ abstract class AbstractSmkRuleOrCheckpointType<T: SmkRuleOrCheckpoint>(
...
+            val displayPath = if (virtualFile == null) {
+                containingFileName
+            } else {
+                val fileContentRootDirectory = ProjectRootManager.getInstance(elem.project)
+                        .fileIndex
+                        .getContentRootForFile(virtualFile)
+                if (fileContentRootDirectory == null) {
+                    containingFileName
+                } else {
+                    VfsUtil.getRelativePath(virtualFile, fileContentRootDirectory)
```
iromeo: 156 could return `null` if the virtual file path doesn't belong to file content root directory, let's return something useful, e.g. virtual file path

Issue #32, opened 2019-02-25
Resolve 'expand' call to snakemake.io


iromeo:
Resolve for 'expand' call to the corresponding place in snakemake.io

(Subtask of #1 )

PR #193, opened 2019-08-14
Quick fixes for several inspections

sharkovadarya:
Issue #194
`SmkRuleRedeclarationInspection` -> rename the second rule.
`SmkResourcesKeywordArgsInspection` -> add `=` and move the caret to the left of it to allow the user to name resource.
`SmkSectionRedeclarationInspection` -> remove the second section redeclaration.
`SmkRuleSectionAfterExecutionInspection` -> move execution section to the end of the rule.


iromeo:
Do we have an issue for this? If yes, please link it, if no, please create a new one.

iromeo:
This pull request changes errors highligting from
image

to 

image

Please revert higlighting it to the previous state

sharkovadarya:
Freshly created issue: 
Inspection quick fixes https://github.com/JetBrains-Research/snakecharm/issues/194

src/main/kotlin/com/jetbrains/snakecharm/inspections/SmkRuleRedeclarationInspection.kt
```diff
@@ -27,8 +29,14 @@ class SmkRuleRedeclarationInspection : SnakemakeInspection() {
...
             if (ruleNames.contains(ruleName)) {
-                registerProblem(rule.nameIdentifier,
-                        SnakemakeBundle.message("INSP.NAME.rule.redeclaration"))
+                val problemElement = rule.nameIdentifier ?: return
+                registerProblem(
+                        problemElement,
+                        SnakemakeBundle.message("INSP.NAME.rule.redeclaration"),
+                        ProblemHighlightType.ERROR,
+                        null,
+                        PyRenameElementQuickFix(problemElement)
```
iromeo: Works weird and produce code which fails on runtime in case of 
```
ruleorder: boo > foo1
localrules: boo

rule boo:
    output: "aa"

rule foo1:
    input: rules.boo

rule boo:
    input: "aa"

rule foo2:
    input: rules.boo
```
Are you sure that it is expected behaviour?
Result is:
```
ruleorder: boo > foo1
localrules: boo

rule boo:
    output: "aa"

rule foo1:
    input: rules.aaaaaaaaaaa

rule aaaaaaaaaaa:
    input: "aa"

rule foo2:
    input: rules.aaaaaaaaaaa
```

src/main/kotlin/com/jetbrains/snakecharm/inspections/SmkResourcesKeywordArgsInspection.kt
```diff
@@ -20,11 +23,48 @@ class SmkResourcesKeywordArgsInspection : SnakemakeInspection() {
...
+                endElement: PsiElement
+        ) {
+            if (!FileModificationService.getInstance().prepareFileForWrite(file)) {
+                return
+            }
+
+            val offset = elementPointer.element?.textOffset ?: return
+            editor?.caretModel?.moveToOffset(offset)
+            if (editor?.document?.isWritable == true) {
+                editor.document.insertString(offset, "=")
```
iromeo: Before
`resources: aaa = "aa", "bb"`
After
`resources: aaa = "aa", <caret>="bb"`

According to my current code style settings should be `resources: aaa = "aa", <caret> = "bb"`, please take in cosideration user code style settings (e.g. `Settings|Editor|Code Style|Python|Spaces|Around Operators|Around = in keyword argument`)

src/main/kotlin/com/jetbrains/snakecharm/inspections/SmkSectionRedeclarationInspection.kt
```diff
@@ -33,12 +36,20 @@ class SmkSectionRedeclarationInspection : SnakemakeInspection() {
                     registerProblem(
                             it,
                             SnakemakeBundle.message("INSP.NAME.section.redeclaration.message", name),
-                            ProblemHighlightType.LIKE_UNUSED_SYMBOL
+                            ProblemHighlightType.LIKE_UNUSED_SYMBOL,
+                            null, RemoveSectionQuickFix()
```
iromeo: Maybe also change section name (w/o updating usages)? E.g.
```
rule foo:
   output: "file.out"
   output: "file.out.log"
```
to 
```
rule foo:
   output: "file.out"
   log: "file.out.log"
```
Such duplication could be a result of copy-paste  + user forgot to update section name

iromeo:
SmkResourcesKeywordArgsInspection could have a similar quick fix as SmkResourcesKeywordArgsInspection for converting positional to keyword args

sharkovadarya:
>SmkResourcesKeywordArgsInspection could have a similar quick fix as SmkResourcesKeywordArgsInspection for converting positional to keyword args

I think there might be a typo?

src/main/kotlin/com/jetbrains/snakecharm/inspections/SmkRuleSectionAfterExecutionInspection.kt
```diff
@@ -50,36 +48,36 @@ class SmkRuleSectionAfterExecutionInspection : SnakemakeInspection() {
...
-        override fun getFamilyName() = SnakemakeBundle.message("INSP.INTN.move.rule.section.upwards.family")
+    private class MoveExecutionSectionToEndOfRuleQuickFix(
+            private val executionSectionPointer: SmartPsiElementPointer<SmkSection>
+    ) : LocalQuickFix {
+        override fun getFamilyName() = SnakemakeBundle.message("INSP.INTN.move.execution.section.down.family")
 
         override fun applyFix(project: Project, descriptor: ProblemDescriptor) {
-            val firstSection = precedingSectionPointer.element ?: return
-            val secondSection = descriptor.psiElement
+            val executionSection = executionSectionPointer.element ?: return
```
iromeo: This `return` means that the user selects the quick fix, presses `apply quick fix` and nothing happens?

sharkovadarya: I guess so. This is to avoid further nullability checks and/or NPE. Normally `element` should not return null, but if in some unforeseeable case it does (documentation: `@return the PSI element, or null if the PSI reparse has completely invalidated the pointer (for example, the element referenced by the pointer has been deleted).`; I am not sure this might happen here, but I am also not sure that it can't ever happen here), what would be a better way to go about it? I don't think displaying an exception to the user is good here, should there be a pop-up message with text "unable to apply quick fix" or something like that?

src/main/kotlin/com/jetbrains/snakecharm/inspections/SmkRuleRedeclarationInspection.kt
sharkovadarya: >Works weird and produce code which fails on runtime in case of
Please clarify what you mean by 'fails at runtime'. I thought  you might mean that there were exceptions but I tried to reproduce this and did not get any, so I'd like to ask for further clarification. 

I assume by 'weird' you mean the fact that the rule got renamed in `rules.` calls but not in localrules and ruleorder sections. I think this might be due to how resolve is implemented for both localrules/ruleorder and `rules.`: if we check usages of the first rule `boo`, we get both localrules/ruleorder and `rules.` usages; for the second rule `boo` we only get `rules`. I suppose this is an issue with resolve, not with this particular action.

It is possible to modify `SmkReferenceExpressionImpl` in a way that renaming of the second rule would also rename its usages in localrules/ruleorder section. However, I assume the desired behavior in this case (that is, for the second rule) is no renaming for usages at all? I am not sure how to implement that unless we abandon multiresolve for `AbstractSmkRuleOrCheckpointType` and `SmkReferenceExpressionImp`, because otherwise even a simple rename action invocation (not in regards to this inspection, just plain old Shift + F6) on the second rule leads to renaming of all its usages. Modifying `SmkReferenceExpressionImpl` is still worth it though, I think, because as of now we probably have an issue with detecting usages?

Anyway, I don't think it's related to this quick fix, but rather to resolve? Unless you were thinking of a completely different problem and I misunderstood your comment.

sharkovadarya: done

src/main/kotlin/com/jetbrains/snakecharm/inspections/SmkResourcesKeywordArgsInspection.kt
sharkovadarya: done

src/main/kotlin/com/jetbrains/snakecharm/inspections/SmkSectionRedeclarationInspection.kt
sharkovadarya: done

src/main/kotlin/com/jetbrains/snakecharm/inspections/SmkRuleSectionAfterExecutionInspection.kt
iromeo: Ok, seems quick fixes doesn't provide API like `isEnabled` / `isValid`.

src/main/kotlin/com/jetbrains/snakecharm/inspections/quickfix/MoveCaretAndInsertEqualsSignQuickFix.kt
```diff
@@ -19,10 +21,13 @@ class MoveCaretAndInsertEqualsSignQuickFix(element: PsiElement) : LocalQuickFixA
...
             endElement: PsiElement
     ) {
+        val pythonCodeStyleSettings = CodeStyle.getCustomSettings(file, PyCodeStyleSettings::class.java)
+        val insertionString = if (pythonCodeStyleSettings.SPACE_AROUND_EQ_IN_KEYWORD_ARGUMENT) " = " else "="
+
         val offset = startElement.textOffset
         editor?.caretModel?.moveToOffset(offset)
         if (editor?.document?.isWritable == true) {
-            editor.document.insertString(offset, "=")
+            editor.document.insertString(offset, insertionString)
```
iromeo: Could we implement it here with 'red box' shown around the place where to type key? E.g. like rule rename quickfix or introduce variable action in python.

src/main/kotlin/com/jetbrains/snakecharm/inspections/SmkRuleSectionAfterExecutionInspection.kt
```diff
@@ -50,36 +48,36 @@ class SmkRuleSectionAfterExecutionInspection : SnakemakeInspection() {
...
             WriteCommandAction.runWriteCommandAction(project) {
-                document.replaceString(startOffset1, endOffset1, text2)
-                if (text1.length > text2.length) {
-                    val endOffset = startOffset1 + text2.length
-                    document.replaceString(startOffset1 + text2.length, endOffset, "")
-                }
-                val startOffset2 = initialStartOffset2 + (text2.length - text1.length)
-                val endOffset2 = initialEndOffset2 + (text2.length - text1.length)
-                document.replaceString(startOffset2, endOffset2, text1)
+                document.deleteString(
```
iromeo: Could you implement it using PsiElement add/delete/replace/etc operations? Smth like in `PyRemoveArgumentQuickFix`?

src/main/kotlin/com/jetbrains/snakecharm/inspections/quickfix/MoveCaretAndInsertEqualsSignQuickFix.kt
sharkovadarya: done

src/main/kotlin/com/jetbrains/snakecharm/inspections/SmkRuleSectionAfterExecutionInspection.kt
sharkovadarya: done

src/main/kotlin/com/jetbrains/snakecharm/inspections/quickfix/IntroduceArgument.kt
```diff
@@ -0,0 +1,50 @@
...
+    override fun invoke(
+            project: Project,
+            file: PsiFile,
+            editor: Editor?,
+            startElement: PsiElement,
+            endElement: PsiElement
+    ) {
+        var element = startElement
+        PyPsiUtils.assertValid(element)
+        if (element.isValid) {
```
iromeo: Seems #31 and #32 line do the same check.

src/main/kotlin/com/jetbrains/snakecharm/inspections/quickfix/IntroduceArgument.kt
```diff
@@ -0,0 +1,50 @@
...
+            val elementGenerator = PyElementGenerator.getInstance(project)
+            val argument = elementGenerator.createKeywordArgument(
+                    LanguageLevel.forElement(element),
+                    defaultArgumentName,
+                    element.text
+            )
+
+            element = element.replace(argument)
+            if (element == null) return
```
iromeo: In which case element will be null here?

src/main/kotlin/com/jetbrains/snakecharm/inspections/quickfix/IntroduceArgument.kt
```diff
@@ -0,0 +1,50 @@
...
+            element = CodeInsightUtilCore.forcePsiPostprocessAndRestoreElement(element)
+            if (element == null) return
+            val builder = TemplateBuilderFactory.getInstance().createTemplateBuilder(element)
+            val keywordArgument = (element as PyKeywordArgument).keywordNode?.psi!!
```
iromeo: Could class cast exception occur here?  `IntroduceArgument()` signature is generic and allow user to use it  for any Psi.

src/main/kotlin/com/jetbrains/snakecharm/inspections/quickfix/IntroduceArgument.kt
```diff
@@ -0,0 +1,50 @@
...
+            builder.replaceElement(keywordArgument, defaultArgumentName)
+            builder.run(editor!!, false)
```
iromeo: The `editor` is nullable according to the method signature, could `editor!!` here result in NPE?

src/main/kotlin/com/jetbrains/snakecharm/inspections/quickfix/IntroduceArgument.kt
sharkovadarya: At this point line `element = element.replace(argument)` has already been executed, and `argument` is created as `PyKeywordArgument`. This `element` is not the original element, it's a keyword argument named `arg` that has the text of the original element for a value. So, if I understand correctly, there will be no class exception; however, if by this comment you mean I should probably make signature more restricting, then I'll change it to `PyExpression`.

iromeo: And if a user enters some invalid text in red box (instead of identification), element will be still PyKeywordArgument?

src/main/kotlin/com/jetbrains/snakecharm/inspections/quickfix/IntroduceArgument.kt
```diff
@@ -0,0 +1,50 @@
...
+                    LanguageLevel.forElement(element),
+                    defaultArgumentName,
+                    element.text
+            )
+
+            element = element.replace(argument)
+            if (element == null) return
+            element = CodeInsightUtilCore.forcePsiPostprocessAndRestoreElement(element)
+            if (element == null) return
+            val builder = TemplateBuilderFactory.getInstance().createTemplateBuilder(element)
```
iromeo: Also not clear `(element as PyKeywordArgument).keywordNode?.psi!!`: it suggest that `keywordNode` is nullable, but finally adds `!!`

src/main/kotlin/com/jetbrains/snakecharm/inspections/quickfix/IntroduceArgument.kt
sharkovadarya: `replace` can return null; however, based on its source code, I can't think of a case when it would return null for `PyKeywordArgument`.

src/main/kotlin/com/jetbrains/snakecharm/inspections/quickfix/IntroduceArgument.kt
sharkovadarya: Documentation: `@param editor the editor where the action is invoked or {@code null} if it's invoked from batch inspections results' tool window.`
Since I don't know what batch inspections are and I wasn't able to find any substantial info on those in the platform docs, I added a null check.

src/main/kotlin/com/jetbrains/snakecharm/inspections/quickfix/IntroduceArgument.kt
sharkovadarya: done

src/main/kotlin/com/jetbrains/snakecharm/inspections/quickfix/IntroduceArgument.kt
sharkovadarya: It won't, but it's not the element in question. The element that gets cast to `PyKeywordArgument` is not the one user performs modifications on (yet). We created a keyword argument from 'arg = somePyExpression'. We want to get a PSI element for `arg` from this freshly created keyword argument. We get this PSI element and we tell `builder` which element should be put in that red box in which the user can enter a new name of their choice. At the moment where the cast is executed, no red box or user input exist yet, only a PSI element which we have created from text and which must be a `PyKeywordArgument` object.

src/main/kotlin/com/jetbrains/snakecharm/inspections/quickfix/IntroduceArgument.kt
iromeo: Ok, then let's add assertion (e.g. requireNonNull) instead of `if (element == null) return` case that shouldn't occur

Issue #200, opened 2019-08-19
F-strings parsing errors in onstart section


iromeo:
It is likely our bug
```
df = []

onstart:
    print(f"Total files: {len(df)}")

rule a:
    output: "out"
```

![image](https://user-images.githubusercontent.com/72933/63283669-8b670180-c2ba-11e9-9520-89ed1c7e56d9.png)

PR #206, opened 2019-08-20
Issue 200

sharkovadarya:
F-strings parsing errors in onstart section https://github.com/JetBrains-Research/snakecharm/issues/200


sharkovadarya:
While this commit does fix this issue, I can only say that the problem in question was due to the fact that `advanceBase()` and `restore(position)` methods had been getting invoked accidentally (after `len` from the example), thus leading to incorrect lexing of the rest of the string (which is not done by us; `tokenText` would return `(df)}")\n\nrule a:\n    output: "out"` instead of `(`; and I'd have to spend considerably more time debugging this in order to figure this out). Now `advanceBase()` is not going to be triggered here, thus not leading to further lexing errors, root cause of which I have been unable to indentify.

Issue #202, opened 2019-08-19
SnakemakeLexer: AIOOBE: Index -1 out of bounds for length 4


iromeo:
```
java.lang.ArrayIndexOutOfBoundsException: Index -1 out of bounds for length 4
	at gnu.trove.TIntArrayList.get(TIntArrayList.java:234)
	at gnu.trove.TIntStack.peek(TIntStack.java:88)
	at com.jetbrains.python.lexer.PythonIndentingProcessor.processIndent(PythonIndentingProcessor.java:379)
	at com.jetbrains.python.lexer.PythonIndentingProcessor.processLineBreak(PythonIndentingProcessor.java:356)
	at com.jetbrains.snakecharm.lang.parser.SnakemakeLexer.processLineBreak(SnakemakeLexer.kt:300)
	at com.jetbrains.python.lexer.PythonIndentingProcessor.processSpecialTokens(PythonIndentingProcessor.java:281)
	at com.jetbrains.python.lexer.PythonIndentingLexer.processSpecialTokens(PythonIndentingLexer.java:18)
	at com.jetbrains.snakecharm.lang.parser.SnakemakeLexer.advance(SnakemakeLexer.kt:191)
	at com.intellij.lang.impl.TokenSequence$Builder.performLexing(TokenSequence.java:86)
	at com.intellij.lang.impl.PsiBuilderImpl.performLexing(PsiBuilderImpl.java:188)
	at com.intellij.lang.impl.PsiBuilderImpl.<init>(PsiBuilderImpl.java:161)
	at com.intellij.lang.impl.PsiBuilderImpl.<init>(PsiBuilderImpl.java:115)
	at com.intellij.lang.impl.PsiBuilderFactoryImpl.createBuilder(PsiBuilderFactoryImpl.java:37)
	at com.jetbrains.python.psi.PyFileElementType.parseContents(PyFileElementType.java:87)
	at com.intellij.psi.impl.source.tree.LazyParseableElement.lambda$ensureParsed$0(LazyParseableElement.java:198)
```

iromeo:
To reproduce, 

1. create Snakefile:
```
import pandas as pd
from pipeline_util import collect_relative_paths

# First generate data_reads.tsv:
# python make_data_table.py config['fastq_dir'] fq.gz data_reads.tsv

df = pd.read_csv("data_reads.tsv", sep='\t')
config["DF"] = df

onstart:
    shell(
        'echo "TMPDIR=$TMPDIR"; '
[...]

PR #208, opened 2019-08-20
Issue 202

sharkovadarya:
SnakemakeLexer: AIOOBE: Index -1 out of bounds for length 4 https://github.com/JetBrains-Research/snakecharm/issues/202


testData/psi/RuleParamsListArgsLineComments.smk
```diff
@@ -65,6 +65,15 @@ rule foo2:
          "echo foo",
         "incorrectly unindented text"
 # end of rule comment
+configfile: "config.yaml"
```
iromeo: Please create a separate test for this case and mention issue202 in test name

PR #216, opened 2019-08-21
Issue 200

sharkovadarya:
F-strings parsing errors in onstart section https://github.com/JetBrains-Research/snakecharm/issues/200

Additionally reworked toplevel/rule section identifying logic to only work on the first tokens on the line or, for rule sections, only if it's immediately after a rule declaration.


src/main/kotlin/com/jetbrains/snakecharm/lang/parser/SnakemakeLexer.kt
```diff
@@ -180,11 +185,7 @@ class SnakemakeLexer : PythonIndentingLexer() {
         }
 
         previousToken = tokenType
-        insideFormattedString = if (!insideFormattedString) {
-            tokenType == PyTokenTypes.FSTRING_START
-        } else {
-            tokenType != PyTokenTypes.FSTRING_END
-        }
+        twoPrecedingTokens = twoPrecedingTokens.copy(first = twoPrecedingTokens.second, second = previousToken)
```
iromeo: Is it a sophisticated way to write `twoPrecedingTokens = twoPrecedingTokens.second to previousToken` ?

Issue #211, opened 2019-08-20
Wrong completion list after rules./checkpoints.


nekit-000000:
![Screenshot from 2019-08-20 18-00-35](https://user-images.githubusercontent.com/25721619/63358925-836b9800-c374-11e9-851f-4d79b839fd66.png)
![Screenshot from 2019-08-20 18-01-36](https://user-images.githubusercontent.com/25721619/63358965-954d3b00-c374-11e9-9681-6536fd3686f2.png)
![Screenshot from 2019-08-20 18-02-18](https://user-images.githubusercontent.com/25721619/63359016-ae55ec00-c374-11e9-81d2-ac4f8bbbb9a3.png)

iromeo:
Fixed

PR #218, opened 2019-08-21
Issue 211

sharkovadarya:
Wrong completion list after rules./checkpoints. https://github.com/JetBrains-Research/snakecharm/issues/211


src/main/kotlin/com/jetbrains/snakecharm/codeInsight/completion/SmkRulesAndCheckpointsCompletionContributor.kt
```diff
@@ -87,9 +87,17 @@ private class SmkRulesAndCheckpointsObjectsCompletionProvider : CompletionProvid
     ) {
         val variants = collectVariantsForElement(parameters.position)
 
+        val parentElement = parameters.position.parent
+        val referencedObject = parameters.withPosition(parentElement, parentElement.textOffset).originalPosition?.parent
+        if (referencedObject is PyReferenceExpression) {
+            when (referencedObject.name) {
+                SnakemakeNames.SMK_VARS_RULES -> variants.removeAll { it.second is SmkCheckPoint }
```
iromeo: Not clear, why we collect both rules and checkpoints from an index in line 88 and then remove rules or checkpoints here. Why not to collect the only required type of elements from the index? Index keys differ for rules and checkpoints.

sharkovadarya: done

src/main/kotlin/com/jetbrains/snakecharm/codeInsight/completion/SmkRulesAndCheckpointsCompletionContributor.kt
```diff
@@ -132,34 +151,24 @@ private class SmkRulesAndCheckpointsObjectsCompletionProvider : CompletionProvid
 
 // the following functions can be used to implement completion for any section/object referring to rule names
 
-private fun collectVariantsForElement(element: PsiElement): MutableList<Pair<String, SmkRuleOrCheckpoint>> {
+private fun <Psi: SmkRuleOrCheckpoint> collectVariantsForElement(
+        element: PsiElement,
```
iromeo: `collectVariantsForElement` is private and is used only here, but has:  too generic and not type-safe signature, e.g. you could call this method with use `indexKey` for rules and pass checkpoints section class. Also its usages introduce code duplication.

Let's use smth simpler, e.g. `collectRuleorCheckpointVariants(element, isCheckPoint=True): List<Pair<String, SmkRuleOrCheckpoint>>`

src/main/kotlin/com/jetbrains/snakecharm/codeInsight/completion/SmkRulesAndCheckpointsCompletionContributor.kt
```diff
@@ -132,34 +151,24 @@ private class SmkRulesAndCheckpointsObjectsCompletionProvider : CompletionProvid
...
+        indexKey: StubIndexKey<String, Psi>,
+        sectionClass: Class<Psi>,
+        currentFileDeclarationsFunction: (SmkFile) -> List<Pair<String, Psi>>
+): MutableList<Pair<String, SmkRuleOrCheckpoint>> {
```
iromeo: Not sure, that usage of `MutableList` here provides noticeable benefits. Immutable collections are more  readable / easy to understand immutable collections. This method force use to think who and why modifies return result of this method, track usages etc.

src/main/kotlin/com/jetbrains/snakecharm/codeInsight/completion/SmkRulesAndCheckpointsCompletionContributor.kt
```diff
@@ -52,7 +53,18 @@ private class SmkRuleNameReferenceCompletionProvider : CompletionProvider<Comple
             context: ProcessingContext,
             result: CompletionResultSet
     ) {
-        val variants = collectVariantsForElement(parameters.position)
+        val variants = collectVariantsForElement(
```
iromeo: Class names for reference providers don't match with implementation, e.g.:
* `SmkRuleNameReferenceCompletionProvider` - adds both rule and checkpoint names into completion
* `SmkRulesAndCheckpointsObjectsCompletionProvider` - adds rule or checkpoint, but not both

src/main/kotlin/com/jetbrains/snakecharm/codeInsight/completion/SmkRulesAndCheckpointsCompletionContributor.kt
sharkovadarya: done

src/main/kotlin/com/jetbrains/snakecharm/codeInsight/completion/SmkRulesAndCheckpointsCompletionContributor.kt
sharkovadarya: done

src/main/kotlin/com/jetbrains/snakecharm/codeInsight/completion/SmkRulesAndCheckpointsCompletionContributor.kt
sharkovadarya: done

src/main/kotlin/com/jetbrains/snakecharm/codeInsight/completion/SmkRulesAndCheckpointsCompletionContributor.kt
```diff
@@ -26,18 +25,18 @@ class SmkRulesAndCheckpointsCompletionContributor : CompletionContributor() {
...
         )
         extend(
                 CompletionType.BASIC,
-                SmkRulesAndCheckpointsObjectsCompletionProvider.IN_SMK_RULES_OR_CHECKPOINTS_OBJECT,
-                SmkRulesAndCheckpointsObjectsCompletionProvider()
+                SmkRulesOrCheckpointsObjectsCompletionProvider.IN_SMK_RULES_OR_CHECKPOINTS_OBJECT,
+                SmkRulesOrCheckpointsObjectsCompletionProvider()
         )
     }
 }
```
iromeo: Here filler name better describes the provider idea. You have 2 providers, one provides completion in `localrules` and `ruleorder` sections, the other one after `rules` and `checkpoints` objects. This information in provider name is easier to read/understand compared to `rules or checkpoints provider` and `rule and checkpoint provider` which look very similar. (no action requied, i've renamed)

src/main/kotlin/com/jetbrains/snakecharm/codeInsight/completion/SmkRulesAndCheckpointsCompletionContributor.kt
```diff
@@ -151,31 +139,38 @@ private class SmkRulesAndCheckpointsObjectsCompletionProvider : CompletionProvid
...
                 .filter { (it as SmkRuleOrCheckpoint).name != null }
                 .map { (it as SmkRuleOrCheckpoint).name!! to it }
         )
     } else {
-        variants.addAll(currentFileDeclarationsFunction.invoke(element.containingFile.originalFile as SmkFile))
+        if (isCheckPoint) {
+            variants.addAll((element.containingFile.originalFile as SmkFile).collectCheckPoints())
+        } else {
+            variants.addAll((element.containingFile.originalFile as SmkFile).collectRules())
+        }
```
iromeo: code duplication here, substing `variants.addAll((element.containingFile.originalFile as SmkFile).`...`)` is quite long and same in lines 163 and 165. (no action required)

Issue #210, opened 2019-08-20
False position arg after keyword arg inspection warning


iromeo:
```
rule bismark_align:
    input:
        #**reads_for_pe_or_se_input({}),
        genome=ancient("indexes/{genome}/{genome}.fa"),  # ensure genome.fa exists
        bismark_indexes_dir=ancient(rules.bismark_indexes.output[0]),
    output:
        bam_file=protected("bams/{sample}/{fq_name}_{genome}_pe.bam"),
        **unaligned_fq()
```
![image](https://user-images.githubusercontent.com/72933/63356512-9a0ff000-c370-11e9-8ba0-8f011cf39ba7.png)

PR #220, opened 2019-08-22
'Positional arg after keyword arg' error for star arguments 

sharkovadarya:
False position arg after keyword arg inspection warning  https://github.com/JetBrains-Research/snakecharm/issues/210

There was a syntax error annotator bug where `positional argument after keyword argument` error would be displayed for unpacked dictionaries or lists, and if you tried to invoke `IntroduceKeywordArgument` quick fix on it, there would be an exception. Star arguments in general shouldn't be highlighted in this case. See `CompatibilityVisitor` class, `highlightIncorrectArguments()` method.


iromeo:
> There was a syntax error annotator bug

Is it about `False position arg after keyword arg inspection warning #210` ?

Issue #184, opened 2019-08-12
Determine wildcard names for rule


iromeo:
Determine which variables in "{foo}" like args are expected to be a wildcard by snakemake. Refer to https://github.com/JetBrains-Research/snakecharm/issues/53#issuecomment-518312488 comment

iromeo:
Be aware of usecases like:

```
rule:
   input: config['reads_dir'] + "/{sample}/{fq_name}{fq_pe_suffix}." + config['fastq_ext']
```

or

```
rule:
   input: f"{config['reads_dir']}/""{sample}/{fq_name}{fq_pe_suffix}." "f{config['fastq_ext']}"
```

or

```
rule:
   input: f"{config['reads_dir']}/{{sample}}/{{fq_name}}{{fq_pe_suffix}}.{config['fastq_ext']}"
```

iromeo:
Also in some sections, other sections names look like wildcards, e.g.
```
rule foo:
  output: "o"
  message:
         "Indexes: Downloading {output}..."
```

iromeo:
Could be used as path related playground
```
rule download_reference_genome_2bit:
    output:
        #twoBit=f"{config['indexes']}/{{genome}}/{{genome}}.2bit")
        #twoBit=config['indexes'] + "/{genome}/{genome}.2bit")
        #twoBit=os.path.join(config['indexes'], "{genome}") + "/{genome}.2bit"
        twoBit=os.path.join(config['indexes'], "{genome}/{genome}.2bit")
    log:
        lambda wd, output: f"{output}.log"
    message:
        "Indexes: Downloading {output}..."
    shell:
        "wget http://hgdownload.soe.ucsc.edu/goldenPath/{wildcards.genome}/bigZips/"
        "{wildcards.genome}.2bit -O {output.twoBit}-o {log}"
```

iromeo:
Here '{output}' is considered is wildcard:

Example
```
rule download_reference_genome_ucsc_2bit:
    output:
        temp("indexes/{genome}/{genome}.2bit")
    wildcard_constraints:
        genome = "hg\\d+"
    log:
        "{output}.log"
        # lambda wd, output: f"{output}.log"
    message:
        "Indexes: Downloading {output}..."
    shell:
        "wget http://hgdownload.soe.ucsc.edu/goldenPath/{wildcards.genome}/bigZips/"
        "{wildcards.genome}.2bit -O {output} -o {log}"
```

SyntaxError:
[...]

PR #225, opened 2019-08-23
#184 determine wildcard names for rule

nekit-000000:
#184


src/main/kotlin/com/jetbrains/snakecharm/lang/psi/SmkWildcardsCollector.kt
```diff
@@ -0,0 +1,50 @@
...
+import com.intellij.psi.util.PsiTreeUtil
+import com.jetbrains.python.psi.PyElementVisitor
+import com.jetbrains.python.psi.PyRecursiveElementVisitor
+import com.jetbrains.python.psi.PyStringLiteralExpression
+import com.jetbrains.snakecharm.stringLanguage.SmkSLFile
+import com.jetbrains.snakecharm.stringLanguage.lang.psi.elementTypes.SmkSLLanguageElement
+import com.jetbrains.snakecharm.stringLanguage.lang.psi.elementTypes.SmkSLReferenceExpressionImpl
+
+class SmkWildcardsCollector : SmkElementVisitor, PyRecursiveElementVisitor() {
+    val collectedWildcards = mutableMapOf<String, PsiElement>()
```
iromeo: Not very good API because exposes mutable map as public field. The only external usage is `collector.collectedWildcards.toList()`

PR #293, opened 2020-07-09
#259 Warn of multiline string arguments

### LINKED ISSUE #259: Warn about multiline string args in input, output, log sections
Likely it is a bug, e.g.
```
rule foo:
  input:
      "file1"
      "file2",
      "file3"
```
Here file1 and file2 will be considered as "file1file2"

SimonTsirikov:
Fixes #259


src/test/resources/features/highlighting/inspections/multiline_string_args_inspetion.feature
```diff
@@ -4,19 +4,38 @@ Feature: Inspection for multiline arguments in same section
...
-      | section    | rule_like  |
-      | input      | rule       |
-      | output     | rule       |
-      | log        | rule       |
-      | input      | checkpoint |
-      | output     | checkpoint |
-      | log        | checkpoint |
\ No newline at end of file
+      | section | rule_like  | args                                |
+      | output  | rule       | "a" \n        "b"                   |
```
iromeo, line 17: Cool, I didn't know that `\n` here will be treated as line break, not just string with `\` and `n` chars! I've checked on runtime, that indeed file text has linebreaks.

src/main/kotlin/com/jetbrains/snakecharm/inspections/SmkSectionMultilineStringArgsInspection.kt
```diff
@@ -16,20 +16,26 @@ class SmkSectionMultilineStringArgsInspection : SnakemakeInspection() {
...
         ) {
 
             val args = argumentList?.arguments ?: emptyArray()
             args.forEach { arg ->
-                if (arg is PyStringLiteralExpression && arg.decodedFragments.size > 1) {
+                if (arg is PyStringLiteralExpression && arg.decodedFragments.size > 1
+                        && arg.stringElements.any { x ->
+                            x.nextSibling is PsiWhiteSpace
+                                    && x.nextSibling.textContains('\n')
+                        }) {
```
iromeo, line 38: Good, by one case isn't covered at all - when argument is string concatenation via '+', I've added and example, see a new failing test.

As one of the possible implementations, you could create a second visitor (e.g. PyElementVisitor) which traverse argument and visits only PyStringLiteralExpression and recursively handles PyBinaryExpression.

SimonTsirikov: Should I make helper visitor class private? It seems unlikely that someone can reuse it.

iromeo: No much sense to make it private

src/main/kotlin/com/jetbrains/snakecharm/inspections/SmkSectionMultilineStringArgsInspection.kt
```diff
@@ -40,20 +42,32 @@ class SmkSectionMultilineStringArgsInspection : SnakemakeInspection() {
...
+    override fun getDisplayName(): String = SnakemakeBundle.message("INSP.NAME.section.multiline.string.args")
+}
+
+class SmkMultilineStringArgsInspectionVisitor : PyElementVisitor() {
+
+    var warning: Boolean = false
+        private set
+
+    override fun visitPyBinaryExpression(node: PyBinaryExpression) {
+        node.leftExpression.accept(this)
```
iromeo, line 56: Please check PyRecursiveElementVisitor for a more elegant and abstract solution for visiting children

src/main/kotlin/com/jetbrains/snakecharm/inspections/SmkSectionMultilineStringArgsInspection.kt
```diff
@@ -31,7 +31,9 @@ class SmkSectionMultilineStringArgsInspection : SnakemakeInspection() {
 
             val args = argumentList?.arguments ?: emptyArray()
             args.forEach { arg ->
-                if (checkRecursive(arg)) {
+                val visitor = SmkMultilineStringArgsInspectionVisitor()
```
iromeo, line 34: Why do we need a separate visitor object here per each section argument? Could we do args visitor stateless and use only a single instance per `SmkSectionMultilineStringArgsInspection` ? E.g. `SmkSectionMultilineStringArgsInspection` is stateless and and just registersProblem in given holder.

SimonTsirikov: One instance of visitor per inspection is enough, as it all use same SnakemakeBundle.message inside one ArgumentList. Now it reports about subexpression in which warning was detected, not whole argument (see test changes in commit bcffa17).

src/main/kotlin/com/jetbrains/snakecharm/inspections/SmkSectionMultilineStringArgsInspection.kt
iromeo: The idea was to check impl of PyRecursiveElementVisitor, not use it. Not inspection is more aggressive and looks recursively even in method calls, etc. E.g.:

image
I don't think that warning for `foo` call here is required. Please add tests for such case.

src/main/kotlin/com/jetbrains/snakecharm/inspections/SmkSectionMultilineStringArgsInspection.kt
iromeo: I was wrong here, we need a psiElement in register problem, so we cannot use a single instance of stateless visitor per inspection. Let's leave current impl here.

src/main/kotlin/com/jetbrains/snakecharm/inspections/SmkSectionMultilineStringArgsInspection.kt
SimonTsirikov: Done

Issue #187, opened 2019-08-13
Lambda expression cannot be used in some rule section


iromeo:
E.g.

```
rule foo:
    output: "{sample}.txt"
    log:  lambda wildcards, output: f"{output}.log"
```
Fails with `'function' object has no attribute 'get_wildcard_names'`. Seems we cannot use lambda as log value, only lambda followed by invocation, e.g. `log: (lambda: ...)()`

Also, lambda in `message` sections fails (`TypeError: expected str, got function
`). Many other sections are likely to not expect lambda function as an argument.

Before implementing please check snakemake sources and documentation. We need to know which sections allows lambda (e.g. input, params).

iromeo:
Is related to #27

sharkovadarya:
Snakemake docs references:
`input`: [input functions](https://snakemake.readthedocs.io/en/stable/tutorial/advanced.html#step-3-input-functions)
`params`: [non-file parameters for rules](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#non-file-parameters-for-rules)
`resources`: [resources](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#resources) (the inspection should be expanded to handle this section)
`threads`: [threads](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#threads) (the inspection should be expanded to handle this section)

sharkovadarya:
Some Snakemake source code references:

`benchmark`, `conda` can't use functions as parameters, see `IOFile` function or `_IOFile` class and its `file()` method in `io.py`.
`output` can't use functions as parameters, see `_set_inoutput_item` method of `Rule` class in `rules.py`.
`log` can't use functions as parameters, see `set_log` and a call to `item.get_wildcard_names()` in `rules.py`, class `Rule`.

sharkovadarya:
Based on these observations, I think `SmkLambdaParameterInSectionInspection` should be extended to handle `threads`, `resources`, and `groups`. There could be another inspection that would highlight usage of lambdas in `message`, `log`, `output`, `conda`, `benchmark`, `singularity`, `shell`, `script`, `wrapper`, `cwl`, `priority` as errors. Usage of lambdas in `wildcard_constraints` and `version` could be highlighted as a warning, although it's highly unlikely that anyone would use lambdas in these functions in the first place.
Generally this extends not only to lambdas, but to any other callables, if I understand correctly.

sharkovadarya:
[...]

PR #305, opened 2020-08-10
#198 warn of unexpected callable args

### LINKED ISSUE #198: Callables can only be used in certain sections
See also: https://github.com/JetBrains-Research/snakecharm/issues/187

Issue 187 only relates to lambda expressions, but similar statements hold true for any sort of callable: `input`, `group`, `resources`, `threads`, `params` can use callables, while other sections cannot.

    def params_function(wildcards, input, output, threads, resources):
        return "text"

    rule foo:
        input: "file.py"
        params: params_function
        shell: "echo hello {params}"

For example, this would work for `params`, but not for `message`.

SimonTsirikov:
Fixes #198


src/main/kotlin/com/jetbrains/snakecharm/inspections/SmkSectionUnexpectedCallableArgsInspection.kt
```diff
@@ -0,0 +1,52 @@
+package com.jetbrains.snakecharm.inspections
```
iromeo, line 1: Please test your feature more attentively, at the moment doesn't work properly:

```
def params_function(wildcards, input, output, threads, resources):
    return "text"

MSG = "msg"
def conda_env(arg):
    return str(arg)

rule foo:
    conda: conda_env # here should be error
    output: MSG # here should be no error, MSG is string, not calable
    benchmark: conda_env("arg") 
    message: MSG   # here should be no error, MSG is string, not calable
    log: params_function  # here should be error
```

image

SimonTsirikov: I found no documentation about which sections accept callable arguments and tested it manually. It seems like in first commit including 'log' section was excess, I removed it now.
Also fixed other references false-positive warning, now it applies only to functions and lambdas.

iromeo: >  But section 'conda' works fine with callable, when I try it locally.
Do you launch snakemake with `--use-conda`? Otherwise, conda sections are just ignored.

iromeo: > I found no documentation about which sections accept callable arguments and tested it manually.

Likely it is only in snakemake sources

iromeo: Callable in `conda` doesn't work without/with `--use-conda`
image

iromeo: > which sections accept callable arguments and tested it manually

As for the list of valid/invalid section names, please see #198 and related source code investigation in #187. Also in snakemake 5.16.0 see the method  `Workflow.rule.decorate` (snakemake/workflow.py:1062). it provides assertions for different section types, e.g. direct check that `theads` and `resources` cannot be a callable. So I believe your test isn't correct.

SimonTsirikov: Found checkpoints for callable arguments in source code of snakemake, now PR fixed according to it.

SimonTsirikov: See [comment](https://github.com/JetBrains-Research/snakecharm/issues/187#issuecomment-674038509) in #187 discussion for extra information.

Issue #260, opened 2019-09-07
Fetch correct wildcards if arg value in rule is link to other section or section arg


iromeo:
Fetch correct wildcards if arg value in rule is link to other section or section arg.

Support required for inspection checking undeclards wildcards and for completion/resolve
```
rule foo1:
   output:
        k = "{sample}.txt"
rule foo2:
   output: 
        "{tag}.txt", 
       "{tag}.svg"
rule boo:
    input:
        rules.foo1.ouput.k, # <- referes 'sample' wildcards
        rules.foo2.output  # <- referes to 'tag' wildcard

**Comment by iromeo** [comment]
Let's assume that in wildcard defining sections ('output/log/benchmark') wildcards defined explicitly, but in sections which could use wildcards (e.g. 'input', 'params', etc) wildcards should be also collected from `rules.*` or `checkpoints.*` links

Subtasks:
- [ ]: Show wildcards in autocompletion, e.g. 'tag' and 'sample' in the rule 'boo'.  
- [ ]: Find usages on wildcard in 'boo' rule should find wildcards in related urls
- [ ]: Update SmkWildcardNotDefinedInspection inspections, see example:
```
rule foo1:
    output:
        out1 = "/path/{sample}_{genome}.1",
        out2 = "/path/{sample}_{genome}.2"

rule foo2:
    output:  "/path/{sample}.txt",

rule foo3:
    output:  "/path/{genome}.txt",

rule second:
    output: "{sample}.txt"
    input:
        "/path/{sample}_{genome}.3", # {genome} wildcard not provided by 'output'
        rules.foo1.output.out1, # [ERROR]: here should be warning that `genome` is missing, like in prev line
        expand(
            rules.foo1.output.out2, genome=["hg19"] # OK: correct
        ),
        rules.foo2.output, # OK
        rules.foo3.output # [ERROR]: {genome} is missing
```
![image](https://user-images.githubusercontent.com/72933/87071713-4b3aa580-c223-11ea-8793-ca4af1f193d7.png)

PR #308, opened 2020-08-11
#260 Error for using reference expression without defining args, when it is necessary

ArtemDavletov:
Error for using reference expression without defining args, when it is necessary.

Fixes: #260

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -12,6 +12,7 @@ Features:
 - Inspection: Comma in the end of section arg list isn't needed (see #255)
 - Inspection: Warn user if rule from 'rulesorder' isn't available in current or included files (see #254)
 - Inspection: Warn about string arguments split on several lines (see #259)
+- Inspection: Error with using reference expression without defining wildcard (see #260)
 
 Fixed:
 - Completion does not suggest top-level keywords in comments (see #178)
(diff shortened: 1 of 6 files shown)


iromeo:
260 issue is only partially implemented, e.g not proper highlighting for case + the only error shown is incorrect

image

```
rule r260_output1:
    output: "foo{genome}{boo}.dd"
    input: "foo{genome}{boo}"

rule r260_output2:
    output: rules.r260_output1.input # {doo} from log not mentioned here
    log: "foo{genome}{doo}.qq.log" # here {boo} from output not mentioned here 
    input: "foo{genome}{doo}" # here {genome} is ok, but {doo} is missing in output
```

Also `boo` not in autocompletion. Please test your impl carefully and think on examples and valid usecases.

iromeo:
Also be aware of possible recursion in rules, it could not have sense on runtime but shouldn't result in Stackoverflow exceptions in plugin.

E.g.:

```
rule r260_wd_recurs1:
    benchmark: rules.r260_wd_recurs2.output

rule r260_wd_recurs2:
    output: rules.r260_wd_recurs3.log

rule r260_wd_recurs3:
    log: rules.r260_wd_recurs1.benchmark
```

Issue #363, opened 2021-05-18
Default path "config/config.yaml" not recognised


iromeo:
Default path "config/config.yaml" not recognised. 

See recommended project layout https://snakemake.readthedocs.io/en/stable/snakefiles/deployment.html

![image](https://user-images.githubusercontent.com/72933/118678757-f3d37180-b805-11eb-9255-b01ca1cf25f6.png)

iromeo:
P.S: in line 3 should be `configfile: "config/config.yaml"` but it is also highlighted as error

iromeo:
See `path_in_workflow_sections_completion.feature`

iromeo:
and `SmkCondaEnvReference`

iromeo:
Demo

![](https://user-images.githubusercontent.com/57076115/125063498-39027a00-e0b8-11eb-997e-f86e47a62cca.gif)

PR #376, opened 2021-07-09
fix: Adding support for default path of configfile #363

KrGolovin:
also adding test for completion and resolve

### DIFF
--- a/src/main/kotlin/com/jetbrains/snakecharm/lang/psi/SmkFileReference.kt
+++ b/src/main/kotlin/com/jetbrains/snakecharm/lang/psi/SmkFileReference.kt
@@ -4,10 +4,7 @@ import com.intellij.codeInsight.lookup.LookupElementBuilder
 import com.intellij.lang.annotation.HighlightSeverity
 import com.intellij.openapi.roots.ProjectRootManager
 import com.intellij.openapi.util.TextRange
-import com.intellij.openapi.vfs.LocalFileSystem
-import com.intellij.openapi.vfs.VfsUtil
-import com.intellij.openapi.vfs.VirtualFile
-import com.intellij.openapi.vfs.VirtualFileManager
+import com.intellij.openapi.vfs.*
 import com.intellij.psi.*
 import com.intellij.psi.util.PsiTreeUtil
 import com.jetbrains.python.psi.PsiReferenceEx
@@ -43,6 +40,7 @@ open class SmkFileReference(
     protected fun collectFileSystemItemLike(
             collectFiles: Boolean = true,
             onlyFromParentFolder: Boolean = true,
+            isFromCurrentDir: Boolean = true,
             predicate: (file: PsiFileSystemItem) -> Boolean = { true }): Array<Any> {
         val project = element.project
 
@@ -63,8 +61,12 @@ open class SmkFileReference(
                 .filter(predicate)
                 .map {
                     val vFile = it.virtualFile
+                    val baseFile = when {
+                        isFromCurrentDir -> parentDirVFile
+                        else -> root
+                    }
                     LookupElementBuilder
-                        .create(relativePath(vFile, parentDirVFile) ?: vFile.path)
+                        .create(relativePath(vFile, baseFile) ?: vFile.path)
                         .withIcon(it.getIcon(0))
                 }
                 .toList()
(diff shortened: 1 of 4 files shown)


KrGolovin:
configfile

src/main/kotlin/com/jetbrains/snakecharm/lang/psi/SmkFileReference.kt
```diff
@@ -158,17 +179,26 @@ class SmkIncludeReference(
...
 ) : SmkFileReference(element, textRange, stringLiteralExpression, path) {
-    override fun getVariants() =  collectFileSystemItemLike() {
+    override fun getVariants() =  collectFileSystemItemLike(
+        onlyFromParentFolder = false,
+        isFromCurrentDir = false
+    ) {
         isYamlFile(it)
     }
+
+    override fun resolve(): PsiElement? = findPathToResolve(searchFromRoot = true)
```
iromeo, line 197: At the moment we have 3 flags: `onlyFromParentFolder`, `isFromCurrentDir`, `searchFromRoot`. At the moment the difference is not clear and no documentation available. I assume that `onlyFromParentFolder`, `isFromCurrentDir` is same thing. Could you refactor this code:
* convert this fields to `SmkFileReference` class abstract variables
* rename them to more clear names + add documentation. I suppose good names could be `searchRelativelyToContentRoot`, `searchRelativelyToCurrentFolder`, where current folder is parent folder of current snakemake file, and content root is the snakemake project working directory

src/main/kotlin/com/jetbrains/snakecharm/lang/psi/SmkFileReference.kt
```diff
@@ -100,28 +102,45 @@ open class SmkFileReference(
...
-    private fun findVirtualFile(): VirtualFile? {
+    private fun findVirtualFile(searchFromRoot: Boolean = false): VirtualFile? {
         val stringLiteral = element.argumentList?.arguments?.firstOrNull {
             it is PyStringLiteralExpression && it.stringValue == path
         } ?: return null
 
-        val parentFolder = element.containingFile.originalFile.virtualFile?.parent ?: return null
-
-        val filePath = (stringLiteral as? PyStringLiteralExpression)?.stringValue
+        val filePath = (stringLiteral as? PyStringLiteralExpression)?.stringValue!!
```
iromeo, line 110: Not clear how it works `(stringLiteral as? PyStringLiteralExpression)?` is marked as nullable, returns `null` or expression, but than is declared as not-null `!!`

iromeo: Moreover prev line always returns `PyStringLiteralExpression` with `stringValue == path`, so there is no need to calculate `stringValue` again. Type checker fails to infer that stringLiteral is `PyStringLiteralExpression` from the above code, but we could refactor it using:

```kotline
        val stringLiteral = element.argumentList?.arguments?.firstOrNull { 
                it is PyStringLiteralExpression &&  it.stringValue == path 
            } as PyStringLiteralExpression? ?: return null

        // or
        val stringLiteral = element.argumentList?.arguments
            ?.filterIsInstance(PyStringLiteralExpression::class.java)
            ?.firstOrNull { it.stringValue == path }
            ?: return null
        
// in both cases `stingLiteral` type will be inferred as `PyStringLiteralExpression`
            
```

iromeo: Also, is not clear why we need `stringLiteral` check in the first line of the method, seems could be refactored just to:
```kotlin
    private fun findVirtualFile(searchFromRoot: Boolean = false): VirtualFile? {
        // Try to find using relative path
        val relativeFile = if (searchFromRoot) {
            //search in all content roots
            ProjectRootManager.getInstance(element.project).contentRoots.firstOrNull { root ->
                root.findFileByRelativePath(path) != null
            }
        } else {
            element.containingFile.originalFile.virtualFile?.parent
                ?.findFileByRelativePath(path)
        }
        if (relativeFile != null) {
            return relativeFile
        }
        // Trying to find the file anywhere
        val vfm = VirtualFileManager.getInstance()
        val localFS = vfm.getFileSystem(LocalFileSystem.PROTOCOL)
        return localFS.findFileByPath(path) ?: vfm.findFileByUrl(path)
    }
```

iromeo (commented):
Feature works, but code needs to be refactored + some tests are missing

src/main/kotlin/com/jetbrains/snakecharm/lang/psi/SmkFileReference.kt
KrGolovin: We have for example "conda:" field,  where we search file from root, but path is making from current folder:
"../envs/snpsift.yaml"
Therefore, we need 2 flags to search for a file and build a path

src/main/kotlin/com/jetbrains/snakecharm/lang/psi/SmkFileReference.kt
KrGolovin: Ok, this is really better, but we have to use firstNotNullOfOrNull

src/main/kotlin/com/jetbrains/snakecharm/lang/psi/SmkFileReference.kt
KrGolovin: I don't understand a bit what an abstract variable is, so for now I have made it the usual fields of the class

iromeo: I meant of abstract property:
```kotlin
abstract class SmkFileReference(..) {
   abstract val  searchRelativelyToCurrentFolder: Boolean
}
class Foo(): SmkFileReference(..) {
  override val searchRelativelyToCurrentFolder = True
} 
```
but your choice (private property + initialization via constructor) is also a good choice. Let's leave it as is.

iromeo: > We have for example "conda:" field
Discussion moved to the new PR.

PR #381, opened 2021-07-12
features/#361-containerized-keyword-support

KrGolovin:
Also adding some tests


KrGolovin:
containerized

src/main/kotlin/com/jetbrains/snakecharm/lang/SnakemakeNames.kt
```diff
@@ -53,6 +54,7 @@ object SnakemakeNames {
     const val SECTION_RUN = "run"
     const val SECTION_CACHE = "cache" // >= 5.12.0
     const val SECTION_CONTAINER = "container"
+    const val SECTION_CONTAINERIZED = "containerized"
```
iromeo, line 57: could you please add a comment in which version the feature was introduced, like for SECTION_CACHER, etc sections here.

src/main/kotlin/com/jetbrains/snakecharm/codeInsight/SnakemakeAPI.kt
```diff
@@ -152,7 +155,8 @@ object SnakemakeAPI {
             SECTION_WRAPPER,
             SECTION_VERSION, SECTION_THREADS,
             SECTION_PRIORITY, SECTION_SINGULARITY, SECTION_CACHE,
-            SECTION_CONTAINER, SECTION_NOTEBOOK, SECTION_ENVMODULES
+            SECTION_CONTAINER, SECTION_CONTAINERIZED, SECTION_NOTEBOOK,
```
iromeo, line 158: Have you checked it on runtime? Is it true, or just an assumption?

KrGolovin: judging by the documentation, containerized has absolutely the same behavior as the container

iromeo: Thx, I've also checked the runtime behaviour and agree with you.

src/main/kotlin/com/jetbrains/snakecharm/lang/SnakemakeNames.kt
KrGolovin: https://github.com/JetBrains-Research/snakecharm/pull/381/commits/f0cbee9bad59cc1aa09c2d78a052ecb2880ba48f

Issue #384, opened 2021-07-13
Some inspections works for rule sections, but not for workflow top-level sections


iromeo:
E.g. not inspection warnings:

* Only one argument is allowed for 'containerized' section. 
* Section 'containerized' does not support keyword arguments 
for 
```
containerized: "docker://continuumio/miniconda3:4.4.10", "ffff", foo="ddd"

rule all:
   input: ""
```

Affected inspections: SmkSectionUnexpectedKeywordArgsInspection, SmkSectionMultipleArgsInspection

dakochik:
Fixed:

![top-level inspections](https://user-images.githubusercontent.com/55440084/126610006-178ac7d5-1dcb-4547-b1e8-2765f00615db.gif)

PR #391, opened 2021-07-22
Features/#384 inspections are enabled for workflows

dakochik:
Issue #384 is resolved


src/main/kotlin/com/jetbrains/snakecharm/inspections/SmkSectionUnexpectedKeywordArgsInspection.kt
```diff
@@ -27,6 +28,12 @@ class SmkSectionUnexpectedKeywordArgsInspection : SnakemakeInspection() {
             }
         }
 
+        override fun visitSmkWorkflowArgsSection(st: SmkWorkflowArgsSection) {
+            if (st.sectionKeyword in SECTIONS_WHERE_KEYWORD_ARGS_PROHIBITED) {
```
iromeo, line 32: You check workflow top-level section name in list of rule/checkpoint section names. Here should be 2 different lists, one for rule/checkpoint set, the other for workflow top level sections. E.g. for 'container' we have SnakemakeNames.WORKFLOW_CONTAINER_KEYWORD and SnakemakeNames.CONTAINER_KEYWORD.

src/main/kotlin/com/jetbrains/snakecharm/inspections/SmkSectionMultipleArgsInspection.kt
```diff
@@ -25,6 +26,13 @@ class SmkSectionMultipleArgsInspection : SnakemakeInspection() {
             }
         }
 
+        override fun visitSmkWorkflowArgsSection(st: SmkWorkflowArgsSection) {
+            val name = st.sectionKeyword
+            if (name != null && name in SINGLE_ARGUMENT_SECTIONS_KEYWORDS) {
```
iromeo, line 31: Same comment as in SmkSectionUnexpectedKeywordArgsInspection

iromeo (changes requested):
Workflow and rule/checkpoint names should be decoupled + please fix also `SmkSyntaxErrorAnnotator` to highlight as error:
```
wildcard_constraints:
    foo = ".*",
    foo = ".*"
```

dakochik:
fixed: highlighting for SmkSyntaxErrorAnnotator
refactored: separate sections for workflows keywords are created

Issue #148, opened 2019-07-25
Unhighlighted syntax error when a function call spans over multiple lines but starts at the same line as section


sharkovadarya:
This code compiles just fine:

    rule foo:
        input: foo1("text1", "text2", "text3")

This, however, produces a compilation error:

    rule foo:
        input: foo1("text1",
                    "text2", "text3")

While this does not:

    rule foo:
        input:
           foo1("text1",
                "text2", "text3")

Would be nice if such an error was highlighted as it is an easy mistake to make but the Snakemake error message (`Expecting rule keyword, comment or docstrings inside a rule definition.`) is not very helpful.

dakochik:
Fixed:

![multiline func call](https://user-images.githubusercontent.com/55440084/126962161-57f37c53-d68d-49e1-9946-a29b200bd2eb.gif)

iromeo:
@dakochik Please add a demo for the new quickfix

dakochik:
New quickfix:

![multiline func call](https://user-images.githubusercontent.com/55440084/127506680-de9c0dbb-3fed-4d9a-8883-d2f13d9a820f.gif)

PR #392, opened 2021-07-26
Features/#148 multiline function call in a single-line declared section

dakochik:
Issue #148 is resolved

### DIFF
--- a/src/main/kotlin/com/jetbrains/snakecharm/inspections/SmkMultilineFunctionCallInspection.kt
+++ b/src/main/kotlin/com/jetbrains/snakecharm/inspections/SmkMultilineFunctionCallInspection.kt
@@ -1,17 +1,16 @@
 package com.jetbrains.snakecharm.inspections
 
 import com.intellij.codeInspection.*
-import com.intellij.lang.ASTNode
-import com.intellij.openapi.command.WriteCommandAction
 import com.intellij.openapi.project.Project
-import com.intellij.psi.PsiElement
-import com.intellij.psi.TokenType
+import com.intellij.psi.*
 import com.intellij.psi.util.elementType
-import com.jetbrains.python.PyElementTypes
+import com.intellij.refactoring.suggested.startOffset
 import com.jetbrains.python.PyTokenTypes
-import com.jetbrains.python.psi.PyArgumentList
+import com.jetbrains.python.psi.PyCallExpression
 import com.jetbrains.snakecharm.SnakemakeBundle
 import com.jetbrains.snakecharm.lang.psi.SmkRuleOrCheckpointArgsSection
+
+const val TAB = "    "
 
 class SmkMultilineFunctionCallInspection : SnakemakeInspection() {
     override fun buildVisitor(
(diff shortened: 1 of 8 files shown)


src/main/kotlin/com/jetbrains/snakecharm/inspections/SmkMultilineFunctionCallInspection.kt
```diff
@@ -0,0 +1,76 @@
...
+        session: LocalInspectionToolSession
+    ) = object : SnakemakeInspectionVisitor(holder, session) {
+        override fun visitSmkRuleOrCheckpointArgsSection(st: SmkRuleOrCheckpointArgsSection) {
+            val args = st.children.firstOrNull { it is PyArgumentList } ?: return
+
+            if (multilineSectionDefinition(args as PyArgumentList)) {
+                return
+            }
+
+            val problemCalls = arrayListOf<PsiElement>()
```
iromeo, line 29: here whitespaces are collected, not PyCallExpressions, so better to rename variable into `invalidWhitespaces`

src/main/kotlin/com/jetbrains/snakecharm/inspections/SmkMultilineFunctionCallInspection.kt
```diff
@@ -0,0 +1,76 @@
...
+            args.children.filter { it.elementType == PyElementTypes.CALL_EXPRESSION }.forEach {
+                multilineFunctionCall(it.node, problemCalls)
+            }
```
iromeo, line 32: let's reduce low level PSI access, e.g. you could filter PyCallExpressions and use API to get arg list:
```kotlin
            args.arguments.forEach { psi ->
                if (psi is PyCallExpression) {
                    collectNewlinesInMultilineCall(psi, invalidWhitespaces) // collectNewlinesInMultilineCall(PyCallExpression, MutableList<PsiElement>)
                }
            }
```

src/main/kotlin/com/jetbrains/snakecharm/inspections/SmkMultilineFunctionCallInspection.kt
```diff
@@ -0,0 +1,76 @@
...
+            element = element.treeNext
+        }
+    }
+
+    private object RemoveSectionQuickFix : LocalQuickFix {
+        override fun getFamilyName() = SnakemakeBundle.message("INSP.NAME.multiline.func.call.fix")
+
+        override fun applyFix(project: Project, descriptor: ProblemDescriptor) {
+            WriteCommandAction.runWriteCommandAction(project) { descriptor.psiElement.delete() }
```
iromeo, line 73: Doesn't work for case:
```
rule rule_148_b:
    input: foo1("text1", #ff
                "text2", "text3")
```

src/main/kotlin/com/jetbrains/snakecharm/inspections/SmkMultilineFunctionCallInspection.kt
```diff
@@ -0,0 +1,76 @@
...
+
+            val problemCalls = arrayListOf<PsiElement>()
+            args.children.filter { it.elementType == PyElementTypes.CALL_EXPRESSION }.forEach {
+                multilineFunctionCall(it.node, problemCalls)
+            }
+
+            problemCalls.forEach {
+                registerProblem(
+                    it,
+                    SnakemakeBundle.message("INSP.NAME.multiline.func.call"),
```
iromeo, line 37: For:
```
rule rule_148_b:
    input: foo1("text1",
                "text2", "text3")
```
2 valid froms available:
```
rule rule_148_b:
    input: foo1("text1","text2","text3")
```
and
```
rule rule_148_b:
    input: 
        foo1("text1",
                "text2", "text3")
```
At the moment the error message forces single line and it could be confusing, if 2nd variant was assumed

iromeo (changes requested):
Not everything works as expected

dakochik:
fixed: quickfix was modified, so now it suggests to transform single-line declared section to multiline declared as follows:
```
# before:
rule rule_148_b:
    input:  foo1("text1",
                "text2", "text3"), foo2("text4")

# after:
rule rule_148_b:
    input: 
        foo1("text1",
            "text2", "text3"),
        foo2("text4")
```

refactored: `multilineSectionDefinition()` was moved into `SmkRuleOrCheckpointArgsSection`, see commit message, some minor changes in inspection code, which were requiered

src/main/kotlin/com/jetbrains/snakecharm/lang/psi/impl/SmkRuleOrCheckpointArgsSectionImpl.kt
```diff
@@ -57,4 +57,24 @@ open class SmkRuleOrCheckpointArgsSectionImpl(node: ASTNode): SmkArgsSectionImpl
...
+
+    override fun multilineSectionDefinition(): Boolean = multilineSectionDefinition(this)
+
+    companion object {
+        fun multilineSectionDefinition(argsSection: SmkArgsSection): Boolean {
+            var node = argsSection.argumentList?.node?.findChildByType(PyTokenTypes.COLON)?.treeNext ?: return false
+            while (true) {
+                node = when (node.elementType) {
+                    TokenType.WHITE_SPACE -> {
+                        if (node.text.contains("\n")) {
```
iromeo, line 69: better to use char instead of single char string `node.text.contains('\n')`

src/main/kotlin/com/jetbrains/snakecharm/inspections/SmkMultilineFunctionCallInspection.kt
```diff
@@ -0,0 +1,98 @@
...
+import com.intellij.openapi.project.Project
+import com.intellij.psi.*
+import com.intellij.psi.util.elementType
+import com.intellij.refactoring.suggested.startOffset
+import com.jetbrains.python.PyTokenTypes
+import com.jetbrains.python.psi.PyCallExpression
+import com.jetbrains.snakecharm.SnakemakeBundle
+import com.jetbrains.snakecharm.lang.psi.SmkRuleOrCheckpointArgsSection
+
+const val TAB = "    "
```
iromeo, line 13: We cannot use such TAB def here, number of spaces in TAB is IDE setting. More over user could use real TAB character instead of spaces.

iromeo: And is only setting for format code features, indents in user file could be different
image

src/main/kotlin/com/jetbrains/snakecharm/inspections/SmkMultilineFunctionCallInspection.kt
```diff
@@ -0,0 +1,98 @@
...
+        override fun invoke(project: Project, file: PsiFile, startElement: PsiElement, endElement: PsiElement) {
+            val doc = PsiDocumentManager.getInstance(project).getDocument(file)
+            if (startElement.prevSibling !is PsiWhiteSpace || doc == null) {
+                return
+            }
+            val argumentList = (startElement as SmkRuleOrCheckpointArgsSection).argumentList ?: return
+            val indent = startElement.prevSibling.text
+            // Moves every argument list element to new line
+            argumentList.arguments.forEach { expression ->
+                doc.insertString(expression.startOffset, "$indent$TAB")
```
iromeo, line 80: I think is better to use `indent` instead of TAB here, because actually you need to indent other lines relative to section name statement

iromeo (changes requested):
Please get rid of harcoded TAB meaning. In general everything works good.

dakochik:
fixed: TAB hardcoding, 'contains' function calling

Issue #355, opened 2021-04-24
Syntax support for Snakemake 6.x 'module' and 'use' concepts


iromeo:
Syntax support for Snakemake 6.x 'module' and 'use' concepts. We need at least basic syntax highlighting.

iromeo:
In current master not everything is parsed properly:

![image](https://user-images.githubusercontent.com/72933/127028983-be12d369-32a4-4bfa-ac81-1673a5352496.png)
![image](https://user-images.githubusercontent.com/72933/127029122-7f433b27-08e4-4952-b2bc-3faa4e3e0ffe.png)

dakochik:
'module' and 'use' sections highlighting is done:

![module use highlighting](https://user-images.githubusercontent.com/55440084/127130323-e88f0c77-b11c-45b1-8edb-9a10b3cc34bc.PNG)

PR #403, opened 2021-07-27
Features/#355 module use support

dakochik:
'module' and 'use' sections highlighting is done

Resolved:  #355


src/main/kotlin/com/jetbrains/snakecharm/lang/parser/SmkStatementParsing.kt
```diff
@@ -289,13 +320,63 @@ class SmkStatementParsing(
...
+        if (!listOfRules && myBuilder.tokenType != PyTokenTypes.IDENTIFIER) { // No rule name, no '*'
+            myBuilder.error(PyPsiBundle.message("PARSE.expected.symbols.first.quotation", "*", "rule name"))
+        } else if (!listOfRules) { // Have rule name, so mark it as reference
+            parseIdentifier()
+        } else { // Have '*'
+            nextToken()
+        }
+
+        if (myBuilder.tokenType != PyTokenTypes.AS_KEYWORD) {
+            checkMatches(PyTokenTypes.FROM_KEYWORD, PyPsiBundle.message("PARSE.0.expected", "from"))
```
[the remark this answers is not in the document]
iromeo: Maybe we need to introduce own token types for 'as', 'with' and remap python lexer token types to our own token types here, because PyCharm shows weird documentation (F1 / mouse hover) that is not relevant.

image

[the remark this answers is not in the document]
iromeo: P.S: we cannot defined 'with' and 'as' in lexer, because it will affect python code parsing, e.g. in cases:
```
with open("fff") as f:
   pass
   
 ```

File: src/main/kotlin/com/jetbrains/snakecharm/lang/parser/SmkStatementParsing.kt [hunk 2]
```diff
@@ -279,23 +310,82 @@ class SmkStatementParsing(
...
+
+            // Creates reference to module definition
+            if (!parseIdentifier(PyElementTypes.REFERENCE_EXPRESSION)) {
```
iromeo, line 342: Why we use Python reference expression, not our own (SmkElementTypes.REFERENCE_EXPRESSION) ?

src/main/kotlin/com/jetbrains/snakecharm/lang/parser/SmkStatementParsing.kt
```diff
@@ -279,23 +310,82 @@ class SmkStatementParsing(
...
+            hasImport = true
+        }
+        // If there are no 'module' keyword, we expect 'as' keyword
+        // If there are 'module' keyword, there may not be 'as' keyword
+        if (!hasImport || myBuilder.tokenType == PyTokenTypes.AS_KEYWORD) {
+            checkMatches(PyTokenTypes.AS_KEYWORD, PyPsiBundle.message("PARSE.0.expected", "as"))
+
+            // New rule name can be: text_*, *_text, text_*_text, text or *
+            val hasFirstIdentifier = myBuilder.tokenType == PyTokenTypes.IDENTIFIER
+            val name = myBuilder.mark()
```
iromeo, line 355: What if `use rule * from rule_355_module as rule_foo` ? Is it runtime error?

src/main/kotlin/com/jetbrains/snakecharm/lang/parser/SmkStatementParsing.kt
```diff
@@ -279,23 +310,82 @@ class SmkStatementParsing(
...
+            }
+
+            hasImport = true
+        }
+        // If there are no 'module' keyword, we expect 'as' keyword
+        // If there are 'module' keyword, there may not be 'as' keyword
+        if (!hasImport || myBuilder.tokenType == PyTokenTypes.AS_KEYWORD) {
+            checkMatches(PyTokenTypes.AS_KEYWORD, PyPsiBundle.message("PARSE.0.expected", "as"))
+
+            // New rule name can be: text_*, *_text, text_*_text, text or *
```
iromeo, line 353: is `text_*_text_*_text` also allowed on runtime?

src/main/kotlin/com/jetbrains/snakecharm/lang/parser/SmkStatementParsing.kt
```diff
@@ -279,23 +310,82 @@ class SmkStatementParsing(
...
+        } else { // Have '*'
+            nextToken()
+        }
+
+        if (myBuilder.tokenType != PyTokenTypes.AS_KEYWORD) {
+            checkMatches(PyTokenTypes.FROM_KEYWORD, PyPsiBundle.message("PARSE.0.expected", "from"))
+
+            // Creates reference to module definition
+            if (!parseIdentifier(PyElementTypes.REFERENCE_EXPRESSION)) {
+                myBuilder.error(PyPsiBundle.message("PARSE.expected.identifier"))
```
iromeo, line 343: Do we have tests for all of this parsing errors? E.g. to ensure that parsing error is shown and text is what we expect

iromeo (changes requested):
In general is good, some case not supported and tests missing. Better to split such huge features (parsing, completion, resolve, inspections) in subtasks and do them sequentially or at least in different commits. One commit should deliver a single idea, not many complex features at once otheriwise it is hard to do code review, check tests coverage, etc.

dakochik:
fixed: 
- parsing logic in commit 479c04a, which affects errors messages, acceptable redeclared rule names, separate tokens for snakemake 'as', 'with' and 'from keywords', references and arguments section parsing. Also this commit adds tests, that check parsing errors.
- commit 8b62434 affects code completion in declaration of use section and adds tests, that checks completion and annotator highlighting.

Answers to some questions:
1) > What if `use rule * from rule_355_module as rule_foo` ? Is it runtime error?

I've checked and found, that snakemake would use as `rule_foo` only the last rule in imported module `rule_355_module`

2) > is `text_*_text_*_text` also allowed on runtime?

Yes :/ 
Commit 479c04a adds support for such names

iromeo (changes requested):
* Not all use syntax support. Maybe is good to look through python impl of use/module parsing in snakemake sources if documentation doesn't cover all cases. 
* Also let's make syntax errors look like snakemake runtime error messages for user convenience (although some of our messages IMHO are better, so let's took best from snakemake runtime errors and our impl).
* List of forbidden sections for `use` seems to be different. Here better is to check snakemake python impl details. E.g. 'benchmark' also is not allowe in use, not only execution sections. Also wrong section name result in weird runtime parsing error, so likely list of supported sections is harcoded in snakemake parser class

src/main/kotlin/com/jetbrains/snakecharm/codeInsight/SnakemakeAPI.kt
```diff
@@ -72,85 +75,89 @@ object SnakemakeAPI {
      * Also see [ImplicitPySymbolsProvider], it also processes 'InputFiles', etc. symbols
      */
     val SECTION_ACCESSOR_CLASSES = mapOf(
-        "snakemake.io.InputFiles" to "input",
-        "snakemake.io.OutputFiles" to "output",
-        "snakemake.io.Params" to "params",
-        "snakemake.io.Log" to "log",
-        "snakemake.io.Resources" to "resources"
```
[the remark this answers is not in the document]
dakochik: Looks like it's a snakemake bug, because i got the semae error in simpler example:
```
use rule name as new_name with:
    output: "file.txt"
    threads: 4
```
But that works fine:
```
use rule name as new_name with:
    output: 
        "file.txt"
    threads: 4
```
The same with your example:
```
use rule rule from other_355_workflow with: # Works fine too
    input: 
        "rule_355.input"
    benchmark: "echo.txt"
```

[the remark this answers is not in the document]
iromeo: Ok, agree, is snakemake bug. 

P.S: In snakemake is parsed using: `...../python3.8/site-packages/snakemake/parser.py`
Block `use` is parsed by `UseRule`, unsupported keywords error is thrown from method `UseRule.block_content(..)`.
```python
    def block_content(self, token):
        if is_comment(token):
            yield "\n", token
            yield token.string, token
        elif is_name(token):
            try:
                self._with_block.extend(self.subautomaton(token.string).consume())
                yield from ()
            except KeyError:
                self.error(
                    "Unexpected keyword {} in rule definition".format(token.string),
                    token,
                )
            except StopAutomaton as e:
                self.indentation(e.token)
                self.block(e.token)
        else:
            self.error(
                "Expecting a keyword or comment "
                "inside a 'use rule ... with:' statement.",
                token,
            )

```

 AFAIU supported sections is taken from `UseRule:  subautomata = rule_property_subautomata`,i.e:
```
rule_property_subautomata = dict(
    name=Name,
    input=Input,
    output=Output,
    params=Params,
    threads=Threads,
    resources=Resources,
    priority=Priority,
    version=Version,
    log=Log,
    message=Message,
    benchmark=Benchmark,
    conda=Conda,
    singularity=Singularity,
    container=Container,
    containerized=Containerized,
    envmodules=EnvModules,
    wildcard_constraints=WildcardConstraints,
    shadow=Shadow,
    group=Group,
    cache=Cache,
    handover=Handover,
)
```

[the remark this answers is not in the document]
iromeo: Thus looks like everything except execution sections. As a result for `shell` we get error like `Unexpected keyword shell in rule definition (rule_355_use.smk, line 21)` , so better error message will be provided by issue https://github.com/JetBrains-Research/snakecharm/issues/408

dakochik:
fixed:
- Support for `use` section declaration with list of imported names see 0c79a6f
- Support for `use` section with single-line declaration see 0c79a6f
- Error messageas are changed and new messages are added see 0c79a6f
- Now parser tries to parse `use` section arguments sections even if there no `with:` see 0c79a6f
- Parser error highlighting tests are changed from `.feature` to `.smk` see 0c79a6f
- `with` completion for (see 5f6371c):
```
use rule rule from other_355_workflow : 
    input: "rule_355.input"
```
- Highlighting `as`, `from`, `with` by smk annotator see 0c79a6f
- Code refactoring in `SmkRuleOrCheckpointNameReference` see 0c79a6f

src/main/kotlin/com/jetbrains/snakecharm/lang/parser/SmkStatementParsing.kt
```diff
@@ -289,13 +318,298 @@ class SmkStatementParsing(
...
+                nextToken()
+                if (myBuilder.tokenType == PyTokenTypes.COMMA) {
+                    myBuilder.error(SnakemakeBundle.message("PARSE.use.wildcard.in.names.list"))
+                    nextToken()
+                    parseIdentifierFromIdentifiersList(names)
+                }
+                marker.done(SmkElementTypes.USE_IMPORTED_RULES_NAMES)
+                asKeywordExists = endOfImportedRulesDeclaration(true, names)
+            }
+            else -> myBuilder.apply {
```
iromeo, line 358: Apply not needed, seems is just `myBuilder.error("...")`

src/main/kotlin/com/jetbrains/snakecharm/lang/parser/SmkStatementParsing.kt
```diff
@@ -289,13 +318,298 @@ class SmkStatementParsing(
...
+                }
+                PyTokenTypes.COMMA -> {
+                    myBuilder.error(SnakemakeBundle.message("PARSE.use.names.expected"))
+                    nextToken()
+                    true
+                }
+                // Actually, Snakemake allows any name for rules and modules,
+                // that doesn't have python token type NAME, which may contains such words as:
+                // 'use', 'as', 'from'
+                // Do we need to add such support?
```
iromeo, line 395: I think we don't need such support now, thx for the comment let's leave it for future =)

src/main/kotlin/com/jetbrains/snakecharm/lang/parser/SmkStatementParsing.kt
```diff
@@ -289,13 +318,298 @@ class SmkStatementParsing(
...
+    }
+
+    /**
+     * Uses when we need to register comma in list of imported rules names
+     * or ending of the list
+     */
+    private fun registerCommaOrEndOfNames(): Boolean {
+        return when (myBuilder.tokenType) {
+            PyTokenTypes.FROM_KEYWORD, PyTokenTypes.AS_KEYWORD, PyTokenTypes.WITH_KEYWORD -> {
+                return false
```
iromeo, line 411: The code looks nasty. We already have 'return when ...' so branches shouldn't use own returns.

src/main/kotlin/com/jetbrains/snakecharm/lang/parser/SmkStatementParsing.kt
```diff
@@ -289,13 +318,298 @@ class SmkStatementParsing(
...
+            }
+        }
+    }
+
+    /**
+     * Uses when we need to register comma in list of imported rules names
+     * or ending of the list
+     */
+    private fun registerCommaOrEndOfNames(): Boolean {
+        return when (myBuilder.tokenType) {
```
iromeo, line 409: Kotlin code style suggests using expression methods instead of single statement method bodies.

src/main/kotlin/com/jetbrains/snakecharm/lang/parser/SmkStatementParsing.kt
```diff
@@ -289,13 +318,298 @@ class SmkStatementParsing(
...
+    /**
+     * Uses when we've finished to collect rules names and went to the next step
+     */
+    private fun endOfImportedRulesDeclaration(
+        definedByWildcard: Boolean,
+        names: List<String>
+    ): Boolean {
+        when (myBuilder.tokenType) {
+            PyTokenTypes.FROM_KEYWORD -> {
+                return fromSignatureParsing(names)
```
iromeo, line 434: Same comment as before 1) return 2) single statement method body

src/main/resources/SnakemakeBundle.properties
```diff
@@ -182,6 +185,22 @@ PARSE.expected.indent=Indent expected
...
+PARSE.use.double.mult.sign=Expecting rulename modifying pattern (e.g. modulename_*) after 'as' keyword
+PARSE.use.requires.wildcard=Multiple rules in 'use rule' statement but name modification ('as' statement) does not contain a wildcard '*'.
+PARSE.use.rule.keyword.expected=Expecting keyword 'rule' after keyword 'use'
+PARSE.use.names.expected=Expecting '*', rule name, or rule listing (comma separated) after 'use rule' statement
+PARSE.use.few.names.from.current.module='use rule' statement from rule in the same module must declare a single rule but multiple rules are declared.
+PARSE.use.unexpected.names.separator=Unexpected token in comma separated list of rules within 'use rule' statement.
+PARSE.use.wildcard.in.names.list=Wildcard '*' can't be used with rules names List
+PARSE.use.unexpected.list.ending=Expecting list of rules in 'use rule' statement to end with keyword 'from'.
+PARSE.use.expecting.module.name=Expecting module name after 'from' keyword in 'use rule' statement.
+PARSE.use.with.missed=':' can be placed only after 'with' keyword, which is missed
```
iromeo, line 198: Some parsing errors ends with '.' here, some not, please using single style. If I'm not mistaken normally we do not use '.' for parsing error messages in python and other languages

iromeo (changes requested):
Almost done =)

dakochik:
fixed: errors messsages, parsing tests
refatored: parser code

Issue #300, opened 2020-07-28
Weak warning for useless log section


dievsky:
Add a weak warning when a useless `log` section is detected. Specifically, when the rule uses `shell` directive, and `log` isn't referenced in the shell command. Maybe it can also be detected in some other cases as well, like with `run`.

Also it could be nice to add a weak warning on a _missing_ log section.

iromeo:
I like the idea. Also could be applicable to `run` section, also harder to do static code analysis

Just a remark: In case of our cluster profile, it uses log section for cluster log even `log` section isn't used by a shell.

iromeo:
E.g.: `unused` log secton:
```
rule:
  log: "foo"
  shell: "echo 123" 
```

ok example:
```
rule:
  log: "foo"
  shell: "echo 123 &> {log}" 
```

dakochik:
Fixed: 

![unused log section](https://user-images.githubusercontent.com/55440084/127613868-5a86b856-41fd-4305-9208-01f54180d6a9.gif)

dakochik:
Updated:

![unused log section](https://user-images.githubusercontent.com/55440084/128136822-609fd58a-df4a-42c7-8291-fd9f92f0829a.gif)

PR #412, opened 2021-07-30
Features/#300 unused log section

dakochik:
Issue #300 is resolved. Now inspection collects all references from SL injections and checks if there any reference to log section, instead of using regular expressions.

It is using `SnakemakeInspectionVisitor` instead of ` SmkSLInspectionVisitor` because in the second case it is less convenient to find and highlight exact unused `log` sections.


src/main/kotlin/com/jetbrains/snakecharm/lang/psi/SmkSLSectionReferencesCollector.kt
```diff
@@ -0,0 +1,81 @@
...
+        subscriptions?.forEach { st ->
+            val child = (st.node.firstChildNode.psi as? SmkSLReferenceExpressionImpl)
+            if (child != null) {
+                checkSLReference(child)
+            }
+        }
+    }
+
+    private fun checkSLReference(ref: SmkSLReferenceExpressionImpl) {
+        if (ref.isWildcard()) {
```
iromeo, line 67: AFAIU you ignore '{log}' where log is wildcard, so isn't considered as log sections usages (that is reasonable), but this case seems not covered by tests

src/main/kotlin/com/jetbrains/snakecharm/lang/psi/SmkSLSectionReferencesCollector.kt
```diff
@@ -0,0 +1,81 @@
...
+import com.jetbrains.python.psi.PyRecursiveElementVisitor
+import com.jetbrains.python.psi.PyStringLiteralExpression
+import com.jetbrains.snakecharm.lang.psi.elementTypes.SmkElementTypes
+import com.jetbrains.snakecharm.stringLanguage.lang.psi.SmkSLFile
+import com.jetbrains.snakecharm.stringLanguage.lang.psi.SmkSLReferenceExpressionImpl
+import com.jetbrains.snakecharm.stringLanguage.lang.psi.SmkSLSubscriptionExpressionImpl
+
+class SmkSLSectionReferencesCollector(
+    private val searchingType: String,
+    private val isRuleOrCheckpointAppropriate: (SmkRuleOrCheckpoint) -> Boolean = { true }
```
iromeo, line 20: For `use` syntax exection section will be missing, so we will likely got a FP on things like
```shell
rule foo:
   shell: "echo {log}"  # inspection warning here could be ignored, it is already reported, no error on runime, undefined sections are empty listrs

use rule foo as boo:
   log: "aa.log"  # <- likely FP here
```
so please add an issue for `use` support to cover uch case in this inspection.

src/main/kotlin/com/jetbrains/snakecharm/inspections/SmkUnusedLogSectionInspection.kt
```diff
@@ -0,0 +1,72 @@
...
+        }
+
+        private fun visitSMKRuleLike(rule: SmkRuleLike<SmkArgsSection>) {
+            val logSection = rule.getSections().firstOrNull {
+                it.sectionKeyword == SnakemakeNames.SECTION_LOG
+            } ?: return
+            val sectionsByRule = session.putUserDataIfAbsent(KEY, hashSetOf())
+
+            if (rule in sectionsByRule) {
+                return
```
iromeo, line 38: Not sure that is actually needed. What is used for? Is it caching that allows next time not to check the rule? What if rule has changed? Inspections will be re-run automatically, but cache could be out-of-data. Not 100% sure, just my guess. Maybe I didn't get the idea.

src/main/kotlin/com/jetbrains/snakecharm/inspections/SmkUnusedLogSectionInspection.kt
```diff
@@ -0,0 +1,72 @@
...
+            }
+
+            val collector = SmkSLSectionReferencesCollector(SnakemakeNames.SECTION_LOG) {
+                !sectionsByRule.contains(rule)
```
iromeo, line 42: Let's rewrite 3 statements as one, like:
```
SmkSLSectionReferencesCollector() {
..
}.also {
  rule.containing.file.accept(it)
}.getSections().foreach() { ...}
```

iromeo, line 42: Why `!sectionsByRule.contains(rule)` doesn't use lambda argument `it` here? is it a bug? If no, why do we need lambda, we could replace this with just boolean value and pass into constructor.

iromeo (changes requested):

src/main/kotlin/com/jetbrains/snakecharm/inspections/SmkUnusedLogSectionInspection.kt
dakochik: As I understand, `sectionsByRule` will be overwritten before each inspection call (in debug mode, `session` saves rules we need while it checks file rules and then, befor next file inspection, it forgots all rules). During one inspection session code can't be modified, so at this point everything should be ok. 
Every time inspection visits rule statement, it scans all available rules, so we use such check in order not to call scanning (via `SmkSLSectionReferencesCollector`) for rules about which we know, that they were referenced during current inspection session

src/main/kotlin/com/jetbrains/snakecharm/lang/psi/SmkSLSectionReferencesCollector.kt
```diff
@@ -0,0 +1,81 @@
...
+            }
+        }
+    }
+
+    private fun checkSLReference(ref: SmkSLReferenceExpressionImpl) {
+        if (ref.isWildcard()) {
+            return
+        }
+        var element = ref.reference.resolve()
+        if (element.elementType == PyElementTypes.KEYWORD_ARGUMENT_EXPRESSION) {
```
iromeo, line 71: Just if `(element is PyKeywordArgument)` seems we don't need low level access to AST node here. ?

src/main/kotlin/com/jetbrains/snakecharm/lang/psi/SmkSLSectionReferencesCollector.kt
```diff
@@ -0,0 +1,81 @@
...
+            element = (element as PyKeywordArgument).parentOfType<SmkRuleOrCheckpointArgsSection>()
+        }
+        if (
+            (element.elementType == SmkElementTypes.RULE_OR_CHECKPOINT_ARGS_SECTION_STATEMENT
```
iromeo, line 75: Same question as before, why do we need AST tree node type access. Seems PSI is enough here.

src/main/kotlin/com/jetbrains/snakecharm/lang/psi/SmkSLSectionReferencesCollector.kt
```diff
@@ -0,0 +1,81 @@
...
+                ?.map { it.first }
+                ?.filterIsInstance<SmkSLFile>()
+                ?: return
+
+        injectedFiles.forEach(::collectResolvedReferences)
+    }
+
+    private fun collectResolvedReferences(file: SmkSLFile) {
+        val references = PsiTreeUtil.getChildrenOfType(file, SmkSLReferenceExpressionImpl::class.java)
+        references?.forEach { st ->
```
iromeo, line 54: Please avoid using `FooImpl`, used corresponding interface `Foo`

iromeo (changes requested):

dakochik:
fixed: now we check if current rule (not any other) refer to `log` section, new quick fixes, additional tests, code refactoring

iromeo (approved):
In general is good, some things could be improved. No action required. PR is accepted.

Issue #413, opened 2021-07-30
Completion features related to 'use' section


dakochik:
- [ ] Add rules, defined in `use` section to `rules.<>` completion list

Related to #355

dakochik:
Fixed:

![use ref](https://user-images.githubusercontent.com/55440084/128729468-6f5658be-7639-4eff-8512-b04ae2001753.PNG)

iromeo:
@dakochik Please add  GIF demo

dakochik:
Done:

![use resolve completion](https://user-images.githubusercontent.com/55440084/129572966-3f467ced-d7e5-4a98-bed3-d43925f07b7a.gif)

Issue #409, opened 2021-07-29
add reference for 'snakefile:' in module declaration


iromeo:
E.g. we could resolve/complete `other_workflow/Snakefile`path  using project *.smk and Snakfile files
```
module other_workflow:
    snakefile: "other_workflow/Snakefile"
```

iromeo:
@dakochik Please add short  GIF demo

dakochik:
Done:

![module file resolve](https://user-images.githubusercontent.com/55440084/129571191-ab44e786-0d21-4c4c-ae1e-3b40b5132832.gif)

PR #415, opened 2021-08-09
Features/#413 use completion

dakochik:
Issue #413 is resolved. Completion and resolve for rule names, which are defined in use section are added. Also it introduces `SmkUnresolvedReferenceInUseSectionInspection` which is used for correct highlighting references to use names, which are used in `use` section definition.


dakochik:
Description of commits, that need to be viwed:

- b01b391 - code refacroting, `SmkUsesType` is removed, `resolve` is done for rules, mentioned in `use` definition and for `rules.<>`. Supports resolve and completion only for current file
- 05bc714 - `resolve` and `completion` in `use` definition and in `rules.<>` now works for imported modules, path to file in `snakefile` sectiion in `module` now is a reference anc can be resolved (see #409 )
- f7565ca - `resolve` and `completion` for rules, which are mentioned in `use` definition or used after `rules.<>` now support rules, which are imported by `include`
- dfc646f - specific highlighting for unresolved rule name reference in `use` definition

dakochik:
Commit 12cd6b9 introduces the same logic as dfc646f but for `rules.<>`. Default python highlighting for unresolved references is enabled.

src/test/resources/features/completion/rules_and_checkpoints_names_completion.feature
```diff
@@ -392,9 +392,17 @@ Feature: Rule and Checkpoints names completion after 'rules.' and 'checkpoints.'
     Given a snakemake project
     And a file "boo.smk" with text
     """
+    include: "zoo.smk"
```
iromeo, line 395: It is better when a test checks a single idea/feature. Here we test several things together - completion from module file and from included file, from undefined module and local rule as other rule. Could you please split test in several independent tests that better reflect which case is being tested.

src/test/resources/features/resolve/rules_and_checkpoints_names_resolve.feature
```diff
@@ -357,12 +357,20 @@ Feature: Resolve name after 'rules.' and 'checkpoints.' to their corresponding d
...
+
     rule something:
       input: "data_file.db"
 
     rule NAME:
       input: "file.txt"
     """
+    And a file "zoo.smk" with text
+    """
+    rule zoo_rule: threads: 1
```
iromeo, line 376: It is better when a test checks a single idea/feature. Here we test several things together - resolve to module and to a file included in module.. Could you please split test in several independent tests that better reflect which case is being tested.

src/test/resources/features/resolve/uses_and_modules_name_resolve.feature
```diff
@@ -27,27 +27,43 @@ Feature: Resolve use and module name to its declaration
...
+    include: "zoo.smk"
+
+    use rule rule_from_zoo as updated_zoo with: threads: 3
+
     rule z:
       input: "data_file.csv"
     """
+    And a file "zoo.smk" with text
+    """
+    rule zoo_rule: threads: 1
```
iromeo, line 43: It is better when a test checks a single idea/feature. Here we test several things together - resolve to module and to a file included in module. Could you please split test in several independent tests that better reflect which case is being tested.

iromeo (changes requested):

src/main/kotlin/com/jetbrains/snakecharm/codeInsight/completion/SmkRulesAndCheckpointsCompletionContributor.kt
```diff
@@ -165,16 +169,25 @@ private fun collectVariantsForElement(
         else -> getVariantsFromIndex(SmkRuleNameIndex.KEY, module, SmkRule::class.java)
     }
 
-    return results.mapNotNull { psi ->
+    val resultList = results.mapNotNull { psi ->
         val name = psi.name
         if (name != null) name to psi else null
     }
+    val usesRules = mutableListOf<Pair<String, SmkRuleOrCheckpoint>>()
+    if (!isCheckPoint) {
```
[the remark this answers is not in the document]
dakochik: It is possible to use `checkpoint` in `use rule` section as you said:
```
rule all:
   input: "file3"
#   run:        # Leads to RE: "'Checkpoints' object has no attribute 'CHECKPOINT_rule'"
#      print(checkpoints.CHECKPOINT_rule.log)        # The same behaviour with common 'rule' instead of 'use rule'

checkpoint CHECKPOINT:
   output: "file2"
   threads: 2
   shell: "touch {output}"

use rule CHECKPOINT as CHECKPOINT_rule with:
   log:
      "mylog"
   output:
      "file3"
   threads: 3
```
So:

1. Yes, I'll fix it
2. No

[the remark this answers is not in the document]
iromeo: > ` print(checkpoints.CHECKPOINT_rule.log)`

Checkpoints usage pattern is slightly different, i.e should be used form input functions and `checkpoints.NAME.get(**wildcards).log`. Likely will lead to the same error, but better to check with correct checkpoints API usage example (you could find checkpoint example in snakemake docs)

dakochik:
Refactored: 

- Mentioned tests were splitted
- Errors messages, inspection decription and unobvious names were modified
- Low-level code was moved into interface methods

fixed: 

- Now it is possible to use `checkpoints` as overriden rules in `use` section definition. Relevant tests are added
- Resolve behaviour now works correctly in cases with wilcards using

src/main/kotlin/com/jetbrains/snakecharm/lang/psi/types/AbstractSmkRuleOrCheckpointType.kt
```diff
@@ -77,6 +81,40 @@ abstract class AbstractSmkRuleOrCheckpointType<T : SmkRuleOrCheckpoint>(
 
     override fun isBuiltin() = false
 
+    protected open fun getUseSections(name: String, location: PyExpression): List<RatedResolveResult> {
+        val result = mutableListOf<RatedResolveResult>()
+        if (location.parent is SmkUse) {
```
iromeo, line 86: could we use `containgRule is SmkUse` here? We already have containing rule field. Not sure that `location.parent` is always containing rule for all use cases.

src/main/kotlin/com/jetbrains/snakecharm/lang/psi/impl/SmkUseImpl.kt
```diff
@@ -29,4 +33,71 @@ class SmkUseImpl : SmkRuleLikeImpl<SmkUseStub, SmkUse, SmkRuleOrCheckpointArgsSe
...
+    }
+
+    override fun getModuleName() =
+        (findChildByType(SmkTokenTypes.SMK_FROM_KEYWORD) as? PsiElement)?.nextSibling?.nextSibling
+
+    override fun getImportedRuleNames() = findChildByType(SmkElementTypes.USE_IMPORTED_RULES_NAMES) as? PsiElement
+
+    override fun getNameIdentifierPattern() = findChildByType(SmkElementTypes.USE_NAME_IDENTIFIER) as? PsiElement
+
+    override fun containsRuleReference(reference: PsiReference) =
```
iromeo, line 101: This method is too specific and is used in a single non-standard case. I think it is ok to inline it and do not provide PSI API for it.

I think `getImportedRuleNames()?.children?` is OK to use in extension and it isn't too low-level as was `node.findChildByType(SmkTokenTypes.SMK_FROM_KEYWORD)?.psi?.nextSibling?.nextSibling` before

src/test/resources/features/resolve/uses_and_modules_name_resolve.feature
```diff
@@ -41,10 +123,10 @@ Feature: Resolve use and module name to its declaration
       input:
         "data_file.txt"
     """
-    When I put the caret at NAME
+    When I put the caret at MODULE as
     Then reference should resolve to "MODULE:" in "foo.smk"
 
-  Scenario: Module name refere to module declaration
+  Scenario: Module name refer to module declaration, confused scenario
```
iromeo, line 129: What does `confused` mean here? Is it for module declaration with unresolved snakefile path?

iromeo: Not clear what is confusing in this scenario compared to previous scenario. But I'm indeed confused =)

iromeo (changes requested):
In general is ok, some tests are missing + completion tests should be refactored + still have couple of questions related to implementation details

src/test/resources/features/resolve/uses_and_modules_name_resolve.feature
```diff
@@ -12,6 +12,10 @@ Feature: Resolve use and module name to its declaration
     """
     When I put the caret at NAME as
     Then reference should resolve to "NAME:" in "foo.smk"
+    Examples:
+      | rule_like  |
+      | rule       |
+      | checkpoint |
```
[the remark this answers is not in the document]
dakochik: Tests are refactored, completion tests (including `checkpoint` references) for `use` declaration statement are added

src/main/kotlin/com/jetbrains/snakecharm/lang/psi/types/AbstractSmkRuleOrCheckpointType.kt
```diff
@@ -77,6 +81,40 @@ abstract class AbstractSmkRuleOrCheckpointType<T : SmkRuleOrCheckpoint>(
 
     override fun isBuiltin() = false
 
+    protected open fun getUseSections(name: String, location: PyExpression): List<RatedResolveResult> {
```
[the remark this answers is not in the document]
dakochik: Done

src/main/kotlin/com/jetbrains/snakecharm/lang/psi/types/AbstractSmkRuleOrCheckpointType.kt
dakochik: I think that it's better to use `location.parent`, because here we want to ignore only the module reference in `use` section, and do not want to ignore e.g. references inside `USE_IMPORTED_RULES_NAMES`, but both references would have `SmkUse` as `containingRule`, so in case of module reference we want to check type of the first parent. Also the `containingRule` is `null` when it is called from `SmkRuleOrCheckpointNameReference` so that's another reason, why I prefere `location.parent`

src/main/kotlin/com/jetbrains/snakecharm/lang/psi/impl/SmkUseImpl.kt
dakochik: Done

src/test/resources/features/resolve/uses_and_modules_name_resolve.feature
dakochik: Done. Changed to:

`Scenario: Module name resolving results shouldn't contain use sections for which the module name matched name pattern`

dakochik:
Identified problems are fixed, see conversations replies

src/main/kotlin/com/jetbrains/snakecharm/lang/psi/types/AbstractSmkRuleOrCheckpointType.kt
iromeo: Ok. I've added a comment with this explanation

iromeo:
@dakochik Please add short  GIF demo

Issue #406, opened 2021-07-27
Support modules and use in statement mover


iromeo:
See `SmkStatementMover`, related to #355

dakochik:
Fixed:

![module mover](https://user-images.githubusercontent.com/55440084/128991348-591297b8-4d8e-4f96-be72-bfa24e473b42.gif)
![use mover](https://user-images.githubusercontent.com/55440084/128991807-2a1c0a3c-5327-4ef6-ae77-cdd59db5fa6b.gif)

PR #416, opened 2021-08-11
Features/#406 modules and use in statement mover

dakochik:
Issue #406 is resolved. `SmkStatementMover` was updated in order to support `use` and `module` sections moving

### DIFF
--- a/src/main/kotlin/com/jetbrains/snakecharm/actions/SmkStatementMover.kt
+++ b/src/main/kotlin/com/jetbrains/snakecharm/actions/SmkStatementMover.kt
@@ -139,6 +139,20 @@ open class SmkStatementMover : PyStatementMover() {
         val statementList = getStatementList(elementToMove) ?: return true
         val statements = statementList.children
 
+        // if statement is a candidate for moving out of statement list:
+        val boundaryStatement = when {
+            down -> statements.last() == elementToMove
+            else -> statements.first() == elementToMove
+        }
+        if (!boundaryStatement) {
+            return true
+        }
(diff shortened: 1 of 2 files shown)


iromeo (changes requested):
1) Exceptions 
2)  doesn't work properly: the issue was before PR, but let's fix it also

Move up:
```
rule NAME:
    output: touch("rule_355/module_355_d.out")
 
use rule NAME as NAME1 with:
    output: touch("rule_355/module_355_d_1.out")
```
results in:
```
rule NAME:
    output: touch("rule_355/module_355_d.out")
    output: touch("rule_355/module_355_d_1.out")

use rule NAME as NAME1 with:
    pass
```
Here rule with single `pass` statement is a syntax error (see https://github.com/JetBrains-Research/snakecharm/issues/420). So we should forbide moving out the latest statement or do not put `pass` in statement list.

dakochik:
I've tried to cancel an insertion of `pass` keyword, but it is inserted even if I delete `addPassStatement` function, that's the reason why I chose the second variant and forbide moving out the latest statement. So now it works better in common cases.

Nevertheless it still doesn't support uncommon cases when we need to add subsection into section with an empty statement list. It used to lead to exception but now the move action is just ignored.
```
rule A:
    intput: "a"
    output: "b" # Move down
rule B:  

output: "b" # Move down
rule C:  
```

iromeo:
>, that's the reason why I chose another variant and forbide moving out the latest statement. 
Ok, sounds reasonable

Issue #408, opened 2021-07-28
Inspection for execution sections in 'use' section


dakochik:
If an execution section appears in 'use' section, we are getting misleading message:

![127199324-845f9c69-9799-4e09-9aab-b8e78e5c5fd8](https://user-images.githubusercontent.com/55440084/127272989-56933eb4-0e85-4157-a229-681cef58a8f4.png)

But in this case, snakemake prints an error with message: "Unexpected keyword shell in rule definition"
So, we need to add a more appropriate message.
Related to #355

iromeo:
Similar for subworflows:
```
subworkflow foo:
    input: ""
```
See `..../python3.8/site-packages/snakemake/parser.py`
```python
        if is_name(token):
            try:
                if token.string == "snakefile":
                    self.has_snakefile = True
                if token.string == "workdir":
                    self.has_workdir = True
                for t in self.subautomaton(token.string).consume():
                    yield t
            except KeyError:
                self.error(
                    "Unexpected keyword {} in "
                    "subworkflow definition".format(token.string),
                    token,
                )
```

iromeo:
For `module`
```
    def block_content(self, token):
        if is_name(token):
            try:
                if token.string == "snakefile":
                    self.has_snakefile = True
                if token.string == "meta_wrapper":
                    self.has_meta_wrapper = True
                for t in self.subautomaton(token.string).consume():
                    yield t
            except KeyError:
                self.error(
                    "Unexpected keyword {} in "
                    "module definition".format(token.string),
                    token,
                )
```

dakochik:
Fixed:

![unrecognized section #1](https://user-images.githubusercontent.com/55440084/128990510-64d924b1-6f42-4acf-880e-467288dc5e00.PNG)
![unrecognized section #2](https://user-images.githubusercontent.com/55440084/128990525-f1fbd968-0488-4173-be6e-b0e51511788c.PNG)

iromeo:
@dakochik Please add short feature demo (png or gif)

dakochik:
Done: 

![unused log section use case](https://user-images.githubusercontent.com/55440084/129573972-8ad7e6b3-73b7-4f29-8c18-4f56e332f01b.PNG)

PR #417, opened 2021-08-11
Features/#408 unexpected sections

dakochik:
Issue #408 is resolved. `SmkUnrecognizedSectionInspection` shows special error message if unrecognized subsection is acceptable for another section.


src/main/kotlin/com/jetbrains/snakecharm/inspections/SmkUnrecognizedSectionInspection.kt
```diff
@@ -46,17 +56,29 @@ class SmkUnrecognizedSectionInspection : SnakemakeInspection() {
...
+                        "INSP.NAME.section.unexpected",
+                        sectionKeyword,
+                        sectionName,
+                        appropriateSection
+                    )
+                }
                 registerProblem(
                     sectionNamePsi,
-                    SnakemakeBundle.message("INSP.NAME.section.unrecognized.message", sectionKeyword),
+                    message,
```
iromeo, line 81: This implementation hardcodes problem type and ignores IDE settings, let's use `ProblemHighlightType.GENERIC_ERROR_OR_WARNING` instead.

src/main/kotlin/com/jetbrains/snakecharm/inspections/SmkUnrecognizedSectionInspection.kt
```diff
@@ -8,11 +8,17 @@ import com.intellij.codeInspection.ui.ListEditForm
...
 import com.jetbrains.snakecharm.lang.psi.elementTypes.SmkElementTypes
 import com.jetbrains.snakecharm.inspections.quickfix.AddIgnoredElementQuickFix
+import com.jetbrains.snakecharm.lang.SnakemakeNames.CHECKPOINT_KEYWORD
+import com.jetbrains.snakecharm.lang.SnakemakeNames.MODULE_KEYWORD
+import com.jetbrains.snakecharm.lang.SnakemakeNames.RULE_KEYWORD
+import com.jetbrains.snakecharm.lang.SnakemakeNames.SUBWORKFLOW_KEYWORD
+import com.jetbrains.snakecharm.lang.SnakemakeNames.USE_KEYWORD
 import javax.swing.JComponent
 
 class SmkUnrecognizedSectionInspection : SnakemakeInspection() {
```
iromeo, line 24: According to the inspection description it highligts section keywords not supported by the plugin. I think #408 should be an independent inspection. More over in case of #408 we are pretty sure that we will get runtime error, thus it should be ERROR, not weak warning.

iromeo (commented):
Works ok, but need to be extracted to a separat inpsection + fix highlighting error type to `ProblemHighlightType.GENERIC_ERROR_OR_WARNING` in order to allow user change inspection highlighting from IDE settings (both inspections)

src/main/kotlin/com/jetbrains/snakecharm/inspections/SmkUnrecognizedSectionInspection.kt
```diff
@@ -72,5 +94,13 @@ class SmkUnrecognizedSectionInspection : SnakemakeInspection() {
 
     override fun createOptionsPanel(): JComponent? =
         ListEditForm(SnakemakeBundle.message("INSP.NAME.section.unrecognized.ignored"), ignoredItems).contentPanel
+
+    private fun getSectionBySubsection(name: String) = when (name) {
+        in RULE_OR_CHECKPOINT_SECTION_KEYWORDS -> RULE_KEYWORD
```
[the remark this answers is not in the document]
dakochik: If I understand your idea correctly, we can't just take section keyword because inspected subsection is already inappropriate for this sectiction and we're using such switch to get the section name, where current subsection can be used

Issue #420, opened 2021-08-11
Do not show syntax error for rule w/o body


iromeo:
```
rule aaa:
pass
```
not an error:
![image](https://user-images.githubusercontent.com/72933/129083175-178e54cd-8925-4e5b-b495-9d1ca4f70ae3.png)

Works on runtime:
```
[Wed Aug 11 21:26:05 2021]
localrule aaa:
    jobid: 0
    resources: tmpdir=/var/folders/qx/grmx87312d3by31f19jvhwgm0000kt/T

[Wed Aug 11 21:26:05 2021]
Finished job 0.
1 of 1 steps (100%) done

```

iromeo:
Also rule with 'pass' statment isn't allowed

```
rule aaa:
    pass
```
results in
```
SyntaxError in line 2 of /Users/romeo/work/snakecharm/snakemaek_examples/untitled10/rule_416.smk:
Unexpected keyword pass in rule definition (rule_416.smk, line 2)
```

iromeo:
Same for `use`:
```

use rule aaa as aa1 with:
    pass

Unexpected keyword pass in rule definition (rule_416.smk, line 5)
```

but
```
use rule aaa as aa1 with:
pass
```
is ok

dakochik:
Looks like snakemake parser adds `pass`  into generated python code by default if `rule` (or `use rule` / `checkpoint`) has an empty body, because for the code:
```
rule NAME1:

rule NAME2:
pass
```

We're getting:
```
@workflow.rule(name='NAME1', lineno=122, snakefile='/mnt/c/Users/Dmitry.Kochik/PycharmProjects/testSMKProj/workflow/Snakefile')
@workflow.norun()
@workflow.run
def __rule_NAME1(input, output, params, wildcards, threads, resources, log, version, rule, conda_env, container_img, singularity_args, use_singularity, env_modules, bench_record, jobid, is_shell, bench_iteration, cleanup_scripts, shadow_dir, edit_notebook, conda_base_path, basedir):
        pass

@workflow.rule(name='NAME2', lineno=128, snakefile='/mnt/c/Users/Dmitry.Kochik/PycharmProjects/testSMKProj/workflow/Snakefile')
@workflow.norun()
@workflow.run
[...]

PR #421, opened 2021-08-13
Features/#420 empty rule statement

dakochik:
Issue #420 is resolved. Now empty `rule` and `checkpoint` are allowed.


src/main/kotlin/com/jetbrains/snakecharm/lang/parser/SmkStatementParsing.kt
```diff
@@ -205,13 +206,22 @@ class SmkStatementParsing(
...
+                nextToken()
                 while (!atToken(PyTokenTypes.DEDENT)) {
                     if (!parseRuleParameter(section)) {
                         break
                     }
                 }
+            } else {
+                if (section.parameterListStatement != SmkElementTypes.RULE_OR_CHECKPOINT_ARGS_SECTION_STATEMENT) {
+                    // Only rules, checkpoints and use rules can have an empty statement list
+                    myBuilder.error(PyPsiBundle.message("indent.expected"))
```
iromeo, line 219: Is `USE_ARGS_SECTION_STATEMENT` missing here?

According to impl this method should be called for `Use` blocks:
```kotlin
tt in SmkTokenTypes.RULE_LIKE -> parseRuleLikeDeclaration(getSectionParsingData(tt!!))
```
and use block expects `SmkElementTypes.USE_ARGS_SECTION_STATEMENT` token type:

```kotlin
   private val useSectionParsingData = SectionParsingData(
        declaration = USE_DECLARATION_STATEMENT,
        name = "use",
        parameterListStatement = SmkElementTypes.USE_ARGS_SECTION_STATEMENT,
        sectionKeyword = SmkTokenTypes.USE_KEYWORD
    )
```

PR #437, opened 2021-11-18
Features/#431 color settings page

dakochik:
Color Settings Page was implemented.


iromeo:
Please add top-level sections examples into demo text. E.g.:
```
configfile: 'config/config.yaml'
localrules: NAME
```
 Not clear is it highlighted like "keyword" or "subsection".

iromeo:
Also try to make the demo shorter. E.g.  3 rules are almost identical. Maybe use one rule with wildcards examples, keyword args, etc. And short example of use rule with one section.

iromeo:
The rest is ok, after fixing demo test merge it into the master.

Issue #109, opened 2019-07-08
Warn user that they need to configure Python sdk with snakemake to enable all features


iromeo:
Warn user that they need to configure Python SDK with snakemake to enable all features. Smart autocompletion & resolve works in `*.smk` files only if Python SDK with snakemake package is configured. This could be not very intuitive for new users.

iromeo:
E.g we could implement `snakemake` facet and detector using `FacetBasedFrameworkDetector`

iromeo:
`ShellcheckSetupNotificationProvider.java` could be useful as example

dakochik:
How to re-enable banner which warms that "Snakemake support is disabled":
* Open `[project_dir]/.idea/snakemake-settings.xml`, where `[project_dir]` is your project directory.
* Search for tag with attribute `smk_support_banner_enabled`.
* Change attribute value to `true` and save changes.

E.g.:
`<component name="SmkProjectSettings" smk_support_banner_enabled="true" />`

PR #443, opened 2021-12-02
Features/#109 sdk config warning

dakochik:
Warning was implemented. Issue #109


src/main/kotlin/com/jetbrains/snakecharm/framework/SmkDisabledBanner.kt
```diff
@@ -0,0 +1,70 @@
...
+import com.intellij.openapi.util.Key
+import com.intellij.openapi.vfs.VirtualFile
+import com.intellij.ui.EditorNotificationPanel
+import com.intellij.ui.EditorNotifications
+import com.jetbrains.snakecharm.SnakemakeBundle
+
+/**
+ * Shows [EditorNotificationPanel] if 'Enable Snakemake support' setting is disabled
+ */
+class SmkDisabledBanner : EditorNotifications.Provider<EditorNotificationPanel>() {
```
iromeo, line 18: Let's rename to SmkSupportDisabledBanner

src/main/kotlin/com/jetbrains/snakecharm/framework/SmkDisabledBanner.kt
```diff
@@ -0,0 +1,70 @@
...
+            EditorNotifications.getInstance(project).updateNotifications(file)
+        }
+        panel.createActionLabel(SnakemakeBundle.message("BANNER.hide")) {
+            // Hides notification for current session
+            hideNotification = true
+            EditorNotifications.getInstance(project).updateNotifications(file)
+        }
+        panel.createActionLabel(SnakemakeBundle.message("BANNER.dont.show.again")) {
+            // Hides notification until cache will be restored
+            PropertiesComponent.getInstance().setValue(DISABLE_NOTIFICATION.toString(), "true")
```
iromeo, line 64: Did you investigated where exactly it is stored and how to re-enable it?

----
In you case you save the setting into Application level component (see implementation in AppPropertiesComponentImpl) so check will be disabled not in current project but in all projects which was or will be opened on user machine. Such behaviour is too strict. More over user doesn't have any option how to undo this choice. According to the source code the option is stored in "other.xml" which is located somewhere in PyCharm config system folders, but user will never find it. 

Better solution is to store it per project, i.e. use API call which takes project as argument and implemented in ProjectPropertiesComponentImpl). In such case user most likely will not find where to re-enable it (property is stored in workspace file stored in `[project_dir].idea/workspace.xml` ) but at least it affects only current project and could be reset it user recreate `.idea' folder, e.g. re-create the project 

So I suggest to:
1) store choice per project, and we have own *.xml file with settings, to which SmkSupportProjectSettings are serialized. So use SmkSupportProjectSettings.
2) Not sure that we need place are UI option for renabling it, but at least we need to describe how to re-enable it in github issue, e.g. that user could change an option in ` .idea/snakemake-settings.xml ` 
3) I don't think that we need an option, which disables the check globally for all projects. But for such cases it could be reasonable to use com.intellij.openapi.util.registry.Registry, which is available in `Tools | Internal Actions | Registry...`. Internal actions menu is shown only in dev instance or if to run PyCharm with additional magic options. This case is slightly better than just storing in some unknown file somewhere in system settingsl

src/main/resources/SnakemakeBundle.properties
```diff
@@ -254,6 +254,12 @@ notifier.msg.framework.by.snakefile.title=Snakemake framework detected
 notifier.msg.framework.by.snakefile.action.configure=Configure Framework...
 notifier.msg.framework.by.snakefile=Snakefile was found in ''{0}''.
 
+#######################
+# Banner
+BANNER.msg=Snakemake support is disabled. Not all features are available.
```
iromeo, line 259: Please be more precisely. `BANNER` is to general. Probably will have other banners in future. E.g use prefix `banner.smk.framework.not.configured'

iromeo (changes requested):
Please store this per-project + add some instruction how to re-enable the check

iromeo:
ok to merge

PR #451, opened 2021-12-18
Features/#359 rules unresolved

dakochik:
The bug cause was the following:
* There are two different methods in `ImplicitPySymbolsProvider`, which collect global variables (e.r. `rules`, `checkpoints`): `addWorkflowGlobalVariables()` and `collectWorkflowGlobalVariables()`. The last one is used only when the `snakemake` version is less than `6.1`, which is checked by `isSmkVersLT_6_1()`.
* `isSmkVersLT_6_1()` method has different behavior depending on whether the program is running in the testing mode or not. That's why in tests everything was good.
* There was a small misspelling in the non test mode case in `isSmkVersLT_6_1()`. Which led to the fact that for any version of the `snakemake`, the method returned `false` (which meant, that `snakemake` had  `6.1` version or higher). So `addWorkflowGlobalVariables()` was always used in non test mode.

Mentioned misspelling was fixed.


iromeo:
Ok to merge.

Issue #452, opened 2021-12-24
Unused log file quick fix


dakochik:
1. Fix behavior with triple quoted string
2. Update fix text e.g. from `shell: "..."` to `shell: "(...) >{log} 2>&1"`

iromeo:
1) Works wrong for string concatenation via '+':

```
    
rule rule_453_4:
    log: "aa"
    shell: 
        "foo boo" +
         "foo boo" +
        "-foo - boo"
 ```
 I suggest here show warning, but do not allow quickfix. Here user could write with '+' just several strings one after another
 
 2) Works incorrect for f-strings, because escaping is required (e.g like `{{log}}`)
```
 rule rule_453_51:
    log: "aa"
    shell:
        "foo boo"
        f"-foo - boo"
 ```
 
 3) Results in syntax errors:
 ```
 rule rule_453_52:
    log: "aa"
    shell:
        f"foo boo"
        "-foo - boo"
        
```

4) For
```
rule rule_453_2:
    log: "aa"
    shell: """
ddd bb
ddd bb
"""
```

will be nice to have:

```
rule rule_453_2:
    log: "aa"
    shell: """(
ddd bb
ddd bb
) >{log} 2>&1 """
```

instead of
```

rule rule_453_2:
    log: "aa"
    shell: """
    (
ddd bb
ddd bb
) >{log} 2>&1 """
```

PR #453, opened 2021-12-24
Features/#452 quick fix in unused log file inspection

dakochik:
#452 Quick fix was fixed.


iromeo:
1) Works wrong for string concatenation via '+':

```
    
rule rule_453_4:
    log: "aa"
    shell: 
        "foo boo" +
         "foo boo" +
        "-foo - boo"
 ```
 I suggest here show warning, but do not allow quickfix. Here user could write with '+' just several strings one after another
 
 2) Works incorrect for f-strings, because escaping is required (e.g like `{{log}}`)
```
 rule rule_453_51:
    log: "aa"
    shell:
        "foo boo"
        f"-foo - boo"
 ```
 
 3) Results in syntax errors:
 ```
 rule rule_453_52:
    log: "aa"
    shell:
        f"foo boo"
        "-foo - boo"
        
```

4) For
```
rule rule_453_2:
    log: "aa"
    shell: """
ddd bb
ddd bb
"""
```

will be nice to have:

```
rule rule_453_2:
    log: "aa"
    shell: """(
ddd bb
ddd bb
) >{log} 2>&1 """
```

instead of
```

rule rule_453_2:
    log: "aa"
    shell: """
    (
ddd bb
ddd bb
) >{log} 2>&1 """
```

dakochik:
Mentioned faults were fixed

dakochik:
As it is written in the 88e1200 commit message, direct string validation was changed to PSI usage

Issue #341, opened 2020-12-14
Inspection: Warn that all docstrings except first will be ignored


iromeo:
```
rule boo:
    'doc-1'
    "doc-2"
    """doc-3"""
    """doc-4"""
    run:
        for rule in workflow.rules:
            print("RULE", rule.name)
            print("DOCSTRING", rule.docstring)
```

Result:
```
$snakemake -j1

RULE boo
DOCSTRING doc-1

```

iromeo:
depends on #114

PR #457, opened 2021-12-30
Features/#341 ignored docstrings inspection

dakochik:
#341 Inspection was implemented.


iromeo:
Seems same could be true for not only for rule/checkpoint, but for subworkflow/use sections. Please investigate it

src/main/kotlin/com/jetbrains/snakecharm/inspections/SmkDocstringsWillBeIgnoredInspection.kt
```diff
@@ -0,0 +1,33 @@
...
+        override fun visitSmkRule(rule: SmkRule) {
+            visitSMKRuleLike(rule)
+        }
+
+        override fun visitSmkCheckPoint(checkPoint: SmkCheckPoint) {
+            visitSMKRuleLike(checkPoint)
+        }
+
+        fun visitSMKRuleLike(rule: SmkRuleLike<SmkArgsSection>) {
+            val docstrings = rule.getStringLiteralExpressions().drop(1)
```
iromeo, line 24: Seems it is a bit complicated way (create new Array list without one element) and more strait-forward  is to check array list length.

iromeo (changes requested):

src/main/kotlin/com/jetbrains/snakecharm/inspections/SmkDocstringsWillBeIgnoredInspection.kt
dakochik: But after that,  I call `registerProblem` for all remaining elements (except the first one, which was dropped). Also I think it's wrong to highlight the first string expression too, because it was placed correctly. If you are agree, we'll need to drop the first element in any case. So, I think there is no other ways to simplify it.

dakochik:
Investigation about subworkflow/use cases:
* Got a RE when I added a docstring in `use rule`. In other cases docstrings are inherited from rules parents.
* There are no docstring element in subworkflow when I run snakemake with `--print-compilation`. Also, I tried `workflow.subworkflows` and then call `docstring` of list's elements. I got the  following error: `'Subworkflow' object has no attribute 'docstring'`. 

So, looks like this inspection can't be applied to subworkflow/use.

src/main/kotlin/com/jetbrains/snakecharm/inspections/SmkDocstringsWillBeIgnoredInspection.kt
iromeo: Ok, let's leave code as is for simplicity. There should be no noticeable performance impact here. 
P.S.: below is just a loop that iterates over docstrings. It could use indexes `1..docstrings.size -1`, coping data to new memory location is not needed for such loop, but it will be wordy. So the current impl is fine here.

Feel free to merge it into master.

iromeo (approved):

Issue #262, opened 2019-09-09
Checkpoints could be accessed as rules when using `rules`


iromeo:
E.g here `foo` is accessed via checkpoints and rules
```
checkpoint foo:
    output:
        key1="{sample}.tsv",
        key2="{sample}.txt"


def boo_input(wildcards):
    path = checkpoints.foo.get(**wildcards).output.key1
    df = pd.read_csv(path, sep="\t")
    return tuple(df[df["mask"]]["File"])


rule boo:
    input: boo_input
    output: "too/{sample}.txt"
    log: "too/{sample}.txt.log"
    shell:
        "echo  {input:q} 1>{output} 2>{log}"

rule all:
    input:
        expand(
            [
                rules.foo.output.key1,
                rules.foo.output.key2,
                rules.boo.output
            ],
            sample=["A", "B"]
        ),
```

PR #459, opened 2021-12-30
Features/#262 checkpoints resolve/completion after rules keyword

dakochik:
Now resolve/completion allow using checkpoints after 'rules' keyword.
#262


iromeo:
@dakochik PR is ok, have you checked the requested behaviour on runtime in snakemake? I'm not 100% sure that it is till true in up-to-date snakemake

dakochik:
When I'm running an example from the issue, I'm getting an RE: `expected string or bytes-like object` in the first `expand` argument.  
But when I'm running more simple example:
```
checkpoint foo:
    output: "{sample}_in"

rule boo:
    input: expand(rules.foo.output, sample=["boo"]) 
    output: "boo_out"
    run:
        shell("touch {output}")
        print(rules.foo.output)
        for rule in workflow.rules: 
            print("RULE",rule.name)
            print(rule.input)
```
I'm getting the following messages:
```
{sample}_in
RULE foo

RULE boo
boo_in
```
And no RE  
  
IDK if it is really necessary feature (I think you are more experienced with snakemake to consider that), but looks like it is still a valid case.

Issue #429, opened 2021-08-24
Show override gutter icon for use / overload rules


iromeo:
Show override gutter icon for use / overload rules

dakochik:
What hasn't been implemented yet:
* Tests
* Probably, we have to improve the way of  overrides collecting.

PR #461, opened 2022-01-13
Features/#429 override gutter icon for use

dakochik:
Gutter icons were added, #429


src/main/kotlin/com/jetbrains/snakecharm/SnakemakeRuleInheritanceMarkerProvider.kt
```diff
@@ -52,17 +54,14 @@ class SnakemakeRuleInheritanceMarkerProvider : RelatedItemLineMarkerProvider() {
         result: MutableCollection<in RelatedItemLineMarkerInfo<*>>
     ) {
         val currentFile = element.containingFile
-        val identifier = (if (element is SmkUse) element.getIdentifierLeaf() else element.nameIdentifier) ?: return
-        val names = if (element is SmkUse) element.getProducedRulesNames().map { it.first } else listOf(
-            element.name ?: return
-        )
-        val e = StubIndex.getInstance().getAllKeys(KEY, element.project)
+        val identifier = (if (element is SmkUse && element.nameIdentifierIsWildcard()) element.nameIdentifier?.firstChild else element.nameIdentifier) ?: return
```
iromeo, line 63: Copy-paste of some complicated logic like in line 34, please use some API instead.

src/main/kotlin/com/jetbrains/snakecharm/SnakemakeRuleInheritanceMarkerProvider.kt
```diff
@@ -31,9 +31,11 @@ class SnakemakeRuleInheritanceMarkerProvider : RelatedItemLineMarkerProvider() {
         use: SmkUse,
         result: MutableCollection<in RelatedItemLineMarkerInfo<*>>
     ) {
-        val name = use.getIdentifierLeaf() ?: return
+        val name = (if (use.nameIdentifierIsWildcard()) use.nameIdentifier?.firstChild else use.nameIdentifier) ?: return
```
iromeo: NameIdentifier is `SmkUseNameIdentifier`. Seems it isn't always SmkUseNameIdentifier, sometimes is just Identifier which inconvenient. But in case of wildcards it is `SmkUseNameIdentifier` on runtime. So please introduce proper PSI api instead of `.firstChild`. Also consider moving `use.nameIdentifierIsWildcard()` to `use.nameIdentifier.isWIldcard()`

At the moment name identifier usage for `use` is a bit complicated and not well structured. Likely some PSI cleanup is requied.

src/main/kotlin/com/jetbrains/snakecharm/SnakemakeRuleInheritanceMarkerProvider.kt
```diff
@@ -31,9 +31,11 @@ class SnakemakeRuleInheritanceMarkerProvider : RelatedItemLineMarkerProvider() {
...
+
         val results =
-            use.getImportedRuleNames()?.mapNotNull { it.reference.resolve() } // Inherited rules are declared in list
+            use.getDefinedReferencesOfImportedRuleNames()
+                ?.mapNotNull { it.reference.resolve() } // Inherited rules are declared in list
```
iromeo, line 44: This call chain is very hard to read and understand. Please refactor it as not a single call chain, so it should be easier to read and understand in which condition what is used.

src/main/kotlin/com/jetbrains/snakecharm/SnakemakeRuleInheritanceMarkerProvider.kt
```diff
@@ -0,0 +1,97 @@
...
+        for (key in e) {
+            files.addAll(StubIndex.getElements(
+                KEY,
+                key,
+                module.project,
+                GlobalSearchScope.moduleWithDependentsScope(module),
+                SmkModule::class.java
+            ).mapNotNull { it.containingFile as? SmkFile })
+        }
+        currentFile.also {
```
iromeo, line 76: Is there any reason of using `also` here? Code could be simplified to just:
```
        if (currentFile is SmkFile) {
            files.add(currentFile)
        }
```
`also` is convenient to use with call chains, see documentation - https://kotlinlang.org/docs/scope-functions.html#also

src/main/kotlin/com/jetbrains/snakecharm/SnakemakeRuleInheritanceMarkerProvider.kt
```diff
@@ -0,0 +1,97 @@
...
+            if (it is SmkFile) {
+                files.add(it)
+            }
+        }
+        val overrides = mutableListOf<SmkRuleOrCheckpoint>()
+        files.forEach { file ->
+            file.collectUses().forEach { pair ->
```
iromeo, line 83: Here code will be easier to understand if instead of `pair -> ... ` you will use unpacked `(useName, psi) -> ..`

iromeo (changes requested):
In general is ok, but some code cleanup is required

dakochik:
Mentioned code cleanup was completed. Additional issues was implemented: #462 #463

Issue #463, opened 2022-01-22
Gutter icon for overriding rules, include case


dakochik:
In the following example:
```
# Snakefile
rule A:
    input: "input_file_A"
```
```
# boo.smk
include: "Snakefile"

use rule A as boo_A with:
    input: "boo_input_file_A"
```
Gutter icon for the rule A was missed.

Issue #462, opened 2022-01-22
Gutter icon for overriding rules, module case


dakochik:
In the following example:
```
# Snakefile
rule A:
    input: "input_file_A"
```
```
# boo.smk
module SNKFILE:
    snakefile: "Snakefile"
```
```
# foo.smk
include: "boo.smk"
use rule A from SNKFILE as foo_A with:
    input: "foo_input_file_A"
```
Gutter icon for the rule A was missed.

PR #465, opened 2022-01-29
Features/#462 gutter icons module case

dakochik:
New index was implemented, so issues #462 and #463 were resolved


dakochik:
Note, that all commits except the las one was taken from #429 so only the fdbdba8 is related to this issue

iromeo (changes requested):
API cleanup required. Please rebase with master first to get PSI changes

dakochik:
API was cleaned-up

PR #466, opened 2022-01-29
Features/#464 module resolve

dakochik:
`module` name resolve was fixed


dakochik:
Note, that all commits except the last one was taken from #462 so only the e890bbd is related to this issue.
Also, sorry, but this commit contains misleading commit message, the following one will be correct:
`module` name resolve behavior was extended by filtering resolve results

iromeo:
Ok, please merge it into master

Issue #515, opened 2024-04-19
Improve parser error message when rule/module is declared with name but lacks ':'


Xewar313:
when declaring rule like:
rule some_identifier
the message pops up saying "Rule name identifier or ':' expected" - this should fix it to only write "':' expected" (since rule cannot have multiple identifiers

PR #516, opened 2024-04-19
Fix/don't mention identifier when it was already defined

Xewar313:
https://github.com/JetBrains-Research/snakecharm/issues/515


iromeo:
* branch name doesn't follow guidelines
* no tests provided
* changelog not updated

Please submit another PR matching the above requirements, at least branch name naming conventions

PR #569, opened 2026-06-26
Declare compatibility with PyCharm/IntelliJ 2026.1 (build 261) — VALIDATED NON-VIABLE, superseded by #570

gaurav:
> ❌ **Closing this PR — validation proves the metadata-only approach does not work on 2026.1.** Superseded by the source port in #570.

## What this tried

PyCharm / IntelliJ **2026.1** is build branch **`261`**, which the released `pluginUntilBuild = 253.*` excludes, so the plugin refuses to load on 2026.1. This PR took the low-risk path: raise the upper bound (`253.* → 261.*`, `pluginVersion → 2025.2.3`) while **still building against PyCharm Community 2025.2** (the last standalone Community release) — i.e. a metadata-only compatibility widening, **no source changes**. The premise was that the unchanged 2025.2 binary would simply *load* on 2026.1.

## Why it does not work (validated)

Ran the IntelliJ Plugin Verifier against **`PY-261.22158.340`** (PyCharm Professional 2026.1). The plugin compiles and the manifest would permit loading, but it hits **4 hard compatibility problems — all from a single class removed in 2026.1**, `com.jetbrains.python.validation.ReturnAnnotator`. Each reference throws **`NoSuchClassError` at runtime**:

```
Plugin SnakeCharm:2025.2.3-eap.SNAPSHOT against PY-261.22158.340: 4 compatibility problems
#Access to unresolved class com.jetbrains.python.validation.ReturnAnnotator
  - SnakemakeVisitorFilter.<init>()                → NoSuchClassError
  - SmkReturnAnnotator.visitPyReturnStatement(...) → NoSuchClassError
  - SmkReturnAnnotator (class)                     → NoSuchClassError
  - SmkReturnAnnotator.<init>()                    → NoSuchClassError
(+ 7 scheduled-for-removal, 4 deprecated incl. PyAnnotator, 155 experimental, 8 internal — not blockers)
```

So raising `pluginUntilBuild` alone would let the plugin **install on 2026.1 and then crash** when the annotator classes load — strictly worse than the current honest "incompatible with this version" rejection. The metadata-only widening is therefore **not viable**; a real source port is required (the `run:`/python-block "return outside function" suppression has to be rebuilt, since `ReturnAnnotator` is gone and the check moved into the `final`, internal `PySyntaxAnnotator`).

[...]

PR #570, opened 2026-06-26
Port SnakeCharm to unified PyCharm 2026.1; the plugin now requires 2026.1 (build 261)

### LINKED ISSUE #575: 135 test failures on a fresh checkout: MockPackages3/snakemake fixture is gitignored and DEVELOPER.md's setup recipe is stale
## Problem

On a fresh checkout of `master`, `./gradlew test` fails **135 Cucumber scenarios** out of 3419. They are not code bugs, and they are not caused by any recent change — they fail identically on plain `master` and on the 2026.1 port branch (#570), where they are easy to misread as port regressions.

Every one of them uses the unversioned `Given a snakemake project` step and fails because the `snakemake` API does not resolve.

## Cause

Two things combine, and neither is visible from the failure output:

1. `testData/MockPackages3/snakemake` is gitignored (`.gitignore:137`), so it is **absent on a clean checkout**. The unversioned `Given a snakemake project` step resolves the snakemake API against it (`src/test/kotlin/features/glue/StepDefs.kt:56-63`), so `resolveQualifiedName("snakemake")` finds nothing, `SmkImplicitPySymbolsProvider` injects nothing, and every dependent reference resolves to 0 targets. The checked-in per-version mocks (`MockPackages3_smk_<ver>`) are unaffected, which is why only some scenarios fail.

2. The setup recipe in `DEVELOPER.md` → Configure Tests, step 2 is stale. It symlinks `~/snakemake/snakemake`, but modern snakemake keeps its package under `src/`, so following the instructions produces a **broken symlink** — the same failures, now with the fixture apparently provisioned.

There is a third trap that makes a *correct* fixture look like it does nothing: the test IDE sandbox persists a VFS/index under `.sandbox_pycharm` that `cleanTest` does not clear. Provisioning the fixture after a previous test run changes nothing until that index is cleared.

## Fix

Verified end-to-end on the **2026.1 port branch** (`update-for-intellij-2026.1`, PY/2026.1.3) — the full suite goes from **135 failures to 1** (3418/3419):

```shell
git clone --branch v9.9.0 https://github.com/snakemake/snakemake.git ~/snakemake
ln -sfn ~/snakemake/src/snakemake testData/MockPackages3/snakemake
find .sandbox_pycharm -maxdepth 3 -name system-test -exec rm -rf {} +
```

The version matters: it must match `snakemake_api.yaml`'s `defaultVersion` (9.9.0), because the FQN tests assert exact 9.9.0 names such as `snakemake.ioutils.subpath.subpath`.

The fixture recovers 134 of the 135. The survivor — `Resolve implicitly imported python names > Warn about unresolved snakemake variable in run section, behaviour differs from scripts` — is a failure the fixture does not address, and it now has its own issue: [#589](https://github.com/JetBrains-Research/snakecharm/issues/589).

**One caveat on that survivor, so nobody inherits an unchecked claim.** It fails on `master` as well as on the port branch, but the measurement above was taken on the port branch, and *failing on both is not evidence of the same cause*. The scenario ends with `When I check highlighting warnings`, which reports only the requested severity and silently discards the rest — the exact trap behind [#584](https://github.com/JetBrains-Research/snakecharm/issues/584), where a warning demoted to `WEAK WARNING` produced the identical `missing (…)` message as one never emitted. #589 carries the diagnosis so far and a one-run recipe to settle whether the two branches fail for the same reason. Nothing else in this issue depends on the answer: the other 134 are unambiguous.

## Proposed resolution

#574 corrects the `DEVELOPER.md` recipe (`src/` layout, version read from `snakemake_api.yaml`, `ln -sfn`, the sandbox-cache gotcha).

Separately, and as a maintainer decision rather than part of that PR: a bare `git clone && ./gradlew test` could be made green with no manual setup by vendoring a trimmed `MockPackages3/snakemake` fixture or automating the clone/symlink in the build. The directory appears to be gitignored deliberately (the full source is large).

### LINKED ISSUE #589: Test failure: 'Warn about unresolved snakemake variable in run section' — the one scenario the fixture does not fix, cause not yet established
`src/test/resources/features/resolve/implicit_py_symbols_resolve.feature:496` fails with the `testData/MockPackages3/snakemake` fixture correctly provisioned. It is the **one** scenario that provisioning the fixture does not fix — the other 134 of the 135 fresh-checkout failures ([#575](https://github.com/JetBrains-Research/snakecharm/issues/575)) all go green.

This issue exists so the failure can be judged on its own rather than as a footnote in a fixture-setup discussion. It supersedes #579, which recorded the 134-vs-135 arithmetic; that correction has since landed in #574 and #575, so #579 is closed.

## The scenario

```gherkin
Scenario: Warn about unresolved snakemake variable in run section, behaviour differs from scripts
  Given a snakemake project
  Given I open a file "foo.smk" with text
   """
   rule a:
       run:
           # this becomes too much and should be migrated into a script directive
           path = snakemake.input[0]
   """
  And PyUnresolvedReferencesInspection inspection is enabled
  Then I expect inspection warning on <input> with message
    """
    Cannot find reference 'input' in '__init__.py'
    """
  When I check highlighting warnings
```

It asserts a **negative**: inside a `run:` block, `snakemake.input` must *not* resolve, unlike in a `script:` file where `snakemake` is the injected job object. Hence "behaviour differs from scripts". The expected message names `__init__.py`, so the assertion only makes sense with the fixture present — `snakemake` resolves to the package, and `input` is then absent from it.

## What is ruled out

**The fixture does not export `input`.** The obvious explanation would be that snakemake v9.9.0 re-exports something named `input` from its package root, making the reference resolve and the expected warning correctly disappear. It does not: `src/snakemake/__init__.py` at `v9.9.0` is 22 lines and re-exports only `shell`.

```python
# snakemake/__init__.py @ v9.9.0, in full apart from dunders and the __main__ guard
import sys
from snakemake._version import version as __version__
PIP_DEPLOYMENTS_PATH = ".snakemake/pip-deployments"
sys.path.append(PIP_DEPLOYMENTS_PATH)
# Reexports that are part of the public API:
from snakemake.shell import shell
```

So `snakemake.input` genuinely is unresolved, and the expected warning *should* fire. Whatever is wrong is on the reporting side, not the fixture.

## What is not yet established

**Whether this fails for the same reason on `master` as on the 2026.1 branch.** #575 states that it "fails on `master` too", but the measurement quoted there (135 → 1, 3418/3419) does not record which branch or platform it ran on, and #579 says the 3-failure figure came from the `update-for-intellij-2026.1` branch. Failing on both is not evidence of a shared cause.

That matters because of [#584](https://github.com/JetBrains-Research/snakecharm/issues/584). This scenario ends with `When I check highlighting warnings`, and that step — `CodeInsightTestFixture.checkHighlighting` — reports only the requested severity plus errors and **silently discards the rest**, so a highlight *demoted* from `WARNING` to `WEAK WARNING` fails with the identical `missing (…)` message as one that is never produced at all. On 2026.1, `PyUnresolvedReferencesInspection` reports through `ProblemHighlightType.LIKE_UNKNOWN_SYMBOL`, which the platform now maps to `HighlightInfoType.INFO` — weak warning — where 2025.2 gave a plain warning. That is exactly the mechanism behind #584, and this scenario uses the same inspection and the same step.

## Next step: one dump run settles it

Add a temporary cucumber step (`OnetimeUsedSteps.kt` is the repo's home for these) that prints what is actually produced, tag the scenario `@here`, and run it:

```kotlin
@When("^dump highlighting$")
fun dumpHighlighting() {
    val infos = SnakemakeWorld.fixture().doHighlighting()
    println("=== HIGHLIGHTING DUMP: ${infos.size} info(s) ===")
    infos.sortedBy { it.startOffset }.forEach { info ->
        println("  ${info.severity} | type=${info.type} | [${info.startOffset}, ${info.endOffset}) | tool=${info.inspectionToolId} | ${info.description}")
    }
    println("=== END HIGHLIGHTING DUMP ===")
}
```

Run it on **both** 2026.1 and `master` (PC/2025.2) and compare severity, message and anchor offset. Three outcomes, with different consequences:

| Dump shows | Diagnosis | Consequence |
| --- | --- | --- |
| Expected message at `WEAK WARNING(200)` | Same signature as #584 — the 2026.1 `LIKE_UNKNOWN_SYMBOL` → `INFO` demotion | **Port-caused on 2026.1**, and #570's "fails identically on `master`" needs correcting |
| Expected message at `WARNING(300)` | The scenario should be passing; the anchor or the message text is what mismatches | Test bug, fix the expectation |
| Message absent entirely | Genuine resolve-side failure | Pre-existing; decide whether to fix the plugin or `@ignore` with this issue linked |

This is the same technique that resolved #584, and it is roughly a minute per branch with the single-feature `@here` recipe — far cheaper than the full suite.

## Notes

- The scenario immediately above this one in the same file — *"Do not warn about unresolved log variable on top-level pipeline handler block"* — is already `@ignore`d for an analogous known divergence. `@ignore` plus a tracking issue is established practice here and may be the right resolution once the cause is known.
- These two are the only scenarios in `implicit_py_symbols_resolve.feature` that enable `PyUnresolvedReferencesInspection`, so there is no passing sibling in the file to use as a control. The two `unresolve_element.feature` scenarios from #584 work as controls instead — they use the same inspection and are known-good on the 2026.1 branch.
- Relevant to [#570](https://github.com/JetBrains-Research/snakecharm/pull/570): that PR is blocked on a confirming full-suite run, and this is the one failure it expects to see. A `cleanTest test` on that branch was cancelled for time at roughly 2,200 of 3,419 tests with **zero** failures recorded, which is consistent with the expected `3419 tests completed, 1 failed` but does not confirm it.

gaurav:
PyCharm 2026.1 ships only as the Professional artifact (`platformType = PY`, build `261.x`) — there is no `pycharm-community:2026.1` — and the Python plugin was rewritten on the way there. This is that port: `PC → PY`, ~37 source-level API fixes, Gradle `8.13 → 9.6.0`, IntelliJ Platform Gradle Plugin `2.7.0 → 2.16.0`, and the test-infrastructure repair needed to get the suite running at all. `pluginSinceBuild` goes `252 → 261`: the binary cannot run on an older IDE, and #569 established by verifier run that advertising otherwise produces "installs, then crashes".

Green end to end — `3419 tests, 0 failures, 0 errors, 0 skipped`, measured on `41c1fb07` (PY/2026.1.3) — plus one scenario added since, the single open item below. `verifyPlugin` reports `Compatible` against `PY-261.27258.51`. Review #573 and #574 first if you can: both are documentation-only and target `master` independently, and both are already merged into this branch, so about a quarter of the insertions here are theirs.

Closes #575
Closes #589

## The calls worth overruling

Most of the diff is forced — the platform changed a signature, the code follows — and the per-break rationale lives in the new [`PORTING.md`](PORTING.md) rather than here. These are the judgement calls:

- **`pluginUntilBuild` stays at `261.*`**, so this build will not load on current PyCharm. 2026.2 is a second port rather than a version bump and lives in #577, stacked on this branch; keeping them apart lets this one be reviewed on its own evidence. `CHANGELOG.md` says so, so users learn it from the release notes rather than the manifest.
- **`PORTING.md` is a new file and cheap to reject** — nothing in the build or source refers to it, and it hangs off two one-line pointers, so dropping it is a two-line revert plus an `rm`.
- **Exactly four test expectations changed, in two feature files, each forced by the platform**: two typeshed stubs that became packages, one message now naming the package rather than the file (#589), two severities the platform demoted (#584). No assertion was weakened — please don't rubber-stamp any expectation change beyond those four. A fifth is *added* (a top-level `yield` must still error), which is a different thing.
- **That demotion is user-visible, and this PR does not decide what to do about it**: an unresolved reference inside an SmkSL injection goes from a yellow warning to a grey one. Restoring the old prominence means raising it to `ERROR` — a product decision, not something to smuggle into a port (#590). Not blocking.
- **Two build-level fixes land here rather than being deferred.** #587's `kotlinx-serialization` alignment: our copy moving to the platform's 1.9.0 is exercised by the 105 wrapper scenarios here, but the `resolutionStrategy` force beside it rests on #577, where it removed 101 failures. And #591: `buildWrappersBundle` now declares its `.cbor` as an output, so `prepareSandbox` wires itself and no hand-written `dependsOn` is left to drop — build-graph surgery inside a platform port, justified only because it *deletes* the wiring this PR twice got wrong. `AGENTS.md` records what must not be tidied up there.

## What's here

**Build** — the bumps above, plus `bundledModule("intellij.spellchecker")` + `bundledPlugin("tanvd.grazi")` now that spellchecker has left core; `verifyPlugin`'s IDE list bound to `pluginSinceBuild`/`pluginUntilBuild` rather than a hardcoded range, which had it verifying six IDEs that can no longer install the plugin; a `[2026.1.0]` changelog section with the unreleased `2025.2.2` one folded in, since only the section matching `pluginVersion` ever renders.

**Source** — `PyType` as a Kotlin interface, the removed `ReturnAnnotator` replaced by a `daemon.highlightInfoFilter` vetoing exactly one message, `CustomFoldingBuilder` nullability.

**Tests** — the suite went from crashing on all 3248 scenarios to running: coroutine `@DebugMetadata` v2, the sandbox layout change, and both `PyTypeShed` helpers-locator crashes from the new v2 content-module packaging (upstream [gradle-plugin #2183](https://github.com/JetBrains/intellij-platform-gradle-plugin/issues/2183), still open).

**Docs** — `AGENTS.md` and a corrected `DEVELOPER.md` arrive via #573/#574, carrying what this port cost to learn.

## Before merging

- [x] **Re-run the full suite at 3420.** The measured figure predates the added `yield` scenario, which is green on its own. Confirmation rather than suspicion, except that scenario isolation here is thinner than it looks (`AGENTS.md` → "Scenario isolation"). ~25 min warm.

Nothing else.

## Not in this PR

2026.2 (#577) · build-wiring tests (#588, and #593 for the `test` half) · #587 on `master`, where the skew is still latent · the 135 environmental failures from the gitignored `MockPackages3/snakemake` fixture (#575, recipe in `DEVELOPER.md`) · the 8 pre-existing `INTERNAL_API_USAGES` that make `verifyPlugin` exit non-zero (#586).


iromeo:
@gaurav Thank you for your recent PR and attempts to fix compatibility with the latest PyCharm releases. Just wanted to let you know, that we see this. Later, I'll review this PR carefully. Due to active work on my research publication, I don't have enough resources right now.

gaurav:
Hi @iromeo, thank you so much for your positive comment! :D Please do wait for me to finish reviewing it myself before reviewing it -- I'll take it off draft and tag you then.

This was pretty much just an experiment for me -- I didn't know anything about IntelliJ plugins or Kotlin, and I was curious to see if Claude Opus 4.8 could figure it out by itself. And after a couple of dead ends when it came up with incorrect explanations for the test failures, I think it's gotten it to work! But of course that means I'd like to review it carefully before submitting it to you to make sure the LLM didn't screw something up. I too have gotten busy in the last few weeks, but I'm hoping to test this out on my local IntelliJ, review the PR, and submit it to you for review next week.

gaurav:
@iromeo I have three small PRs ready for you to review:
* https://github.com/JetBrains-Research/snakecharm/pull/576
* https://github.com/JetBrains-Research/snakecharm/pull/574
* https://github.com/JetBrains-Research/snakecharm/pull/572

As you can see @claude did most of the work (I've [created a PR](https://github.com/JetBrains-Research/snakecharm/pull/573) to add an AGENTS.md file to the repo if you'd like to encourage coding agents!) but I've reviewed all three of those PRs and then look good to me. But please let me know if you'd like me to change anything, or feel free to change them yourself.

gaurav:
@iromeo This PR is now passing all tests! My Kotlin skills are not strong, so I'm mostly trusting that Claude was able to get all the tests passing without any huge changes to the system that I noticed, but you might spot issues that I didn't -- if so, please feel free to let me know what needs fixing. Worst case, having this PR (and especially PORTING.md, which documents that changes that had to be made to get the test passing) might help you write a cleaner update to get Snakecharm working on PyCharm 2026.1.

Note that there are a few PRs have been merged into this one (since I can't stack PRs from a fork onto a parent repo) -- I would recommend reviewing those PRs first, letting me know if things need improving, and then once those are merged I/you can rebase this PR onto `main` so you can see only the changes needed to get PyCharm 2026.1 to work.

Issue #571, opened 2026-06-26
Hardcoded local snakemakeWrappersRepoPath in gradle.properties breaks buildPlugin for other contributors


gaurav:
## Summary

`gradle.properties` ships a **hardcoded, machine-specific** default for `snakemakeWrappersRepoPath` that points at a JetBrains developer's home directory:

https://github.com/JetBrains-Research/snakecharm/blob/aa529e6b82ac4226244caa94a879a5a4ebafa8db/gradle.properties#L109

```properties
snakemakeWrappersRepoPath=/Users/romeo/work/git_repos/snakecharm_related/snakemake-wrappers
```

Because this path doesn't exist on anyone else's machine, `./gradlew buildPlugin` fails for outside contributors at `:buildWrappersBundle` (which runs `SmkWrapperCrawler` with this path):

```
Execution failed for task ':buildWrappersBundle'.
> Process 'command '.../bin/java'' finished with non-zero exit value 1
```

`./gradlew test` is unaffected (it uses `buildTestWrappersBundle` with the in-repo `testData/wrappers_storage`), so only the distributable-ZIP path is broken.

## Suggestion

Make the default contributor-friendly, e.g. one of:
- default `snakemakeWrappersRepoPath` to a repo-relative location (or leave it unset and have `buildWrappersBundle` skip/emit an empty bundle when it's missing), and document the `-PsnakemakeWrappersRepoPath=<path>` override; and/or
- document in `DEVELOPER.md` that `buildPlugin` requires a local `snakemake-wrappers` checkout (tag matching `snakemakeWrappersRepoVersion`, currently `v7.2.0`) with the path overridden.

There's also a commented machine-specific path at [`gradle.properties#L65`](https://github.com/JetBrains-Research/snakecharm/blob/aa529e6b82ac4226244caa94a879a5a4ebafa8db/gradle.properties#L65) (`#platformLocalPath=/Users/romeo/...`) that could be cleaned up at the same time.

_Found while building the plugin locally to test 2026.1 compatibility (see #569 / #570)._

PR #573, opened 2026-07-16
Add AGENTS.md, a build/test/architecture guide for coding agents

gaurav:
Adds a top-level `AGENTS.md`: a tool-agnostic orientation file for AI coding agents (and human newcomers) working in this repo. It complements, and defers to, `DEVELOPER.md` and `README.md` rather than duplicating them — the aim is the handful of things that cost hours to discover and are not written down anywhere.

Everything in it was learned the hard way while working on #570, #572, #574 and #577, and every claim was re-checked against the tree, not against memory of the session that produced it.

### Contents

- **What the plugin is** — an IntelliJ Platform plugin layered on the bundled Python plugin's PSI/API (most EPs registered against `language="Python"`).

- **Build** — the JDK 21 toolchain, including two traps: the pinned Gradle crashes under a much newer JVM with `Type T not present`, and on macOS `/usr/libexec/java_home -v 21` treats 21 as a *minimum*, so it happily returns a newer JDK and exits 0. Also the Kotlin-daemon OOM workaround, and what `snakemakeWrappersRepoPath` does and does not gate now that #572 has made it optional (issue #571): it fronts `buildPlugin` and `runIde`, but **not** the test tasks, which route through `prepareTestSandbox` and the separate test-only bundle.

- **Running tests** — the recipes that turn a 25-minute suite into a 60-second one and stop a stale run from lying to you: the `@here` tag plus the `AllCucumberFeaturesTest` filter; `cleanTest` because `testData` is not a declared input of `test`; that scenarios share one cached project whenever they hand the light-fixture framework the same `LightProjectDescriptor`, so a step must set the project state it needs rather than assume a fresh default; how to triage a wall of failures out of `build/test-results/test/TEST-*.xml`; and the fresh-checkout `testData/MockPackages3/snakemake` fixture whose absence costs ~135 spurious failures (#574).

- **Architecture** — the two languages (Snakemake driving the Python `PyParser` API; SmkSL injected into strings), snakemake versions as language levels via `snakemake_api.yaml`, the feature-area → package → `features/` test-dir mapping, and the "grep `plugin.xml` to find a feature's entry class" tip.

- **Platform conventions** — `gradle.properties` as the source of truth, the `YEAR.MAJOR.MINOR` scheme and its `CHANGELOG.md` requirement, that 2025.2 is the last standalone PyCharm Community release, that a platform bump also moves the Kotlin compiler / Java toolchain / `intelliJPlatform` baselines (each failing before your source is even considered), that `TestLoggerFactory` turns one logged error into hundreds of failed scenarios, and the post-2025.2 modular-platform gotcha requiring explicit `bundledModule`/`bundledPlugin` declarations.

## Notes for reviewers

- **Is an `AGENTS.md` welcome here at all?** That is the real question this PR asks, which is why it is deliberately separate from any feature work — it can be declined without affecting anything else.
- Naming is a choice: `AGENTS.md` is the emerging cross-tool convention, so the content lives there once. Claude Code is the exception — it reads `CLAUDE.md` — so this PR also adds a one-line `CLAUDE.md` containing `@AGENTS.md`, an import rather than a copy, so there is nothing to keep in sync. Happy to rename either file, or fold the content into `DEVELOPER.md`, if you would rather have one file.
- If you disagree with a claim in it, that is a bug worth filing — the file is only useful if it is exactly right.

Related: #574 (the fixture fix behind the "fresh-checkout gotcha" section), #570 / #577 (the platform ports that produced the bump notes), #572 (wrappers path, merged), #585 (build-side follow-ups found while checking this file's wrappers claims).

<details>
<summary><b>Review history</b> — kept for anyone tracing why a particular line reads the way it does; the durable conclusions are all in the file above.</summary>

- The wrappers paragraph originally claimed `buildWrappersBundle` "sits in front of `buildPlugin`, `runIde` **and** the test tasks". It does not: only `prepareSandbox` depends on it, and `test --dry-run` never lists it. Corrected, and the same wrong claim in a `build.gradle.kts` comment is fixed in #585.
- The shared-test-project paragraph first only alluded to the trap; it now says what actually causes it (same `LightProjectDescriptor` → same cached project) and gives `Given a snakemake with disabled framework project` as the cautionary example.
- Reviewing this file's build claims turned up three real build-script problems — a silent wrapper-less publish on CI, an empty property taking the quiet path, and a stale bundle that could be packed — none of which belong in a docs PR. They are #585.

</details>

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -37,12 +37,13 @@ PyCharm Professional. Note that **2025.2 is the last standalone PyCharm Communit
 
 **Wrappers bundle:** `:buildWrappersBundle` reads `snakemakeWrappersRepoPath` (a local
 [snakemake-wrappers](https://github.com/snakemake/snakemake-wrappers) checkout) and runs as part of
-`prepareSandbox`, so it sits in front of `buildPlugin`, `runIde` **and** the test tasks. The property
-`snakemakeWrappersRepoPath` is unset by default in `gradle.properties` so the plugin bundle and local IDE
-run will not provide wrappers related completion and other features until the variable is set in properties
-or using cmdline like `./gradlew buildPlugin -PsnakemakeWrappersRepoPath=/path/to/snakemake-wrappers`. 
-The test-only bundle (`:buildTestWrappersBundle`, what `test` actually consumes) reads by default
-`testData/wrappers_storage` and needs no property. 
+`prepareSandbox`, so it sits in front of `buildPlugin`, `runIde` **and** the test tasks. That
+property is commented out in `gradle.properties` by default, so a plain `buildPlugin` / `runIde`
+yields a plugin without wrapper completion and the other wrapper-driven features; pass it explicitly
+to include them: `./gradlew buildPlugin -PsnakemakeWrappersRepoPath=/path/to/snakemake-wrappers` (on
+TeamCity it comes from the wrappers VCS root — see issue #571). The test-only bundle
+(`:buildTestWrappersBundle`, what `test` actually consumes) defaults to `testData/wrappers_storage`
+and needs no property.
 
 **CLI build memory:** if `:compileKotlin` dies with `OutOfMemoryError: GC overhead limit exceeded`,
 give the Kotlin daemon more heap — append `-Pkotlin.daemon.jvmargs=-Xmx4g` (transforming some large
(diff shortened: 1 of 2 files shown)


iromeo:
Thanks, looks good! I guess we could merge it. There is mention of #574 PR, so after #574 will be merged the file will required some cleanup

gaurav:
Yeah, this is WIP because I'm updating this file as I work on the other PRs -- once all my PRs are done, I'll review this manually and then submit for review. But if you want to merge this now, I'm also happy to update AGENTS.md within those PRs -- whatever is easiest for you!

iromeo:
>  But if you want to merge this now, I'm also happy to update AGENTS.md within those PRs 
For me it is not critical merge it now or later. To reduce number of PR I'd rather merge it now and update later within PRs

gaurav:
Sweet -- let me polish this up a bit on Monday, and then I'll submit it to you for your review.

gaurav:
@iromeo Ready for your review!

PR #574, opened 2026-07-18
Improve stale test-fixture setup docs

### LINKED ISSUE #575: 135 test failures on a fresh checkout: MockPackages3/snakemake fixture is gitignored and DEVELOPER.md's setup recipe is stale

gaurav:
On a fresh checkout, `./gradlew test` fails **135 Cucumber scenarios** (of 3419 — 3248 scenarios plus 171 plain JUnit tests). They are not code bugs. They are a missing test fixture, and the setup instructions in `DEVELOPER.md` are stale in three ways that each make it look like something else. This PR fixes those instructions. **134 of the 135 failures go away**; the survivor is a genuine pre-existing failure that `master` has too.

The same 135 appear in the 2026.1 port ([#570](https://github.com/JetBrains-Research/snakecharm/pull/570)), where they read as port regressions. They are not — they fail identically on plain `master`.

Fixes #575.

## What's here

One file: **`DEVELOPER.md`**, in two places.

**1. Configure Tests, step 2 — the fixture recipe.** The unversioned `Given a snakemake project` scenarios resolve the snakemake API against `testData/MockPackages3/snakemake`, which is gitignored and therefore absent on a clean checkout. Four things were missing or wrong in the instructions for providing it:

- **`src/` layout.** Modern snakemake keeps its package under `src/`, so the symlink target is `~/snakemake/src/snakemake`. The old `~/snakemake/snakemake` path creates a **broken symlink** — the failures persist, now with the fixture apparently provisioned. The recipe uses `ln -sfn` because anyone who followed the old instructions already has that broken symlink, and a plain `ln -s` aborts with "File exists" and leaves it there.
- **Version read from `snakemake_api.yaml`.** The fixture must match `defaultVersion` (9.9.0 today), which the FQN tests assert exact names against, e.g. `snakemake.ioutils.subpath.subpath`. The recipe reads it out of the file rather than hardcoding it, so the instructions survive a `defaultVersion` bump instead of going stale the same way again.
- **Re-runnable clone.** Anyone who followed the old recipe already has `~/snakemake`, so a plain `git clone … ~/snakemake` aborts for exactly the readers the previous point caters to — and with the commands unchained, `ln -sfn` still ran and pointed the fixture at whatever revision that clone happened to sit on. It now clones only when the directory is absent, then checks out the tag, chained so a bad version cannot reach the symlink.
- **The "zero effect" gotcha.** The test IDE sandbox persists a VFS/index under `.sandbox_pycharm` that **`cleanTest` does not clear**. Add the fixture after having run the tests once and the stale index never sees it, so the failures persist unchanged — which is how a correctly-provisioned fixture looks like it did nothing. Clearing it is an unconditional part of the fix, not a caveat: the fixture is gitignored, so clean CI never has it either. The documented command uses `find` rather than a glob, because the sandbox sits at different depths depending on how tests were launched and on the platform-plugin version, and a glob that misses deletes nothing while appearing to succeed.

**2. A new "Reading test results" section.** It starts from what `./gradlew test` already gives you — a line per failing scenario, then `N tests completed, M failed` — notes that HTML reports are disabled in `build.gradle.kts` (Windows can't handle some Cucumber scenario names) so the JUnit XML under `build/test-results/test/` is what a finished run leaves on disk, and gives a `tee` + `grep`/`sort`/`diff` recipe for the one thing the console alone won't do: compare two runs, to see which failures a change fixed or introduced.

## What it produces

Verified end to end on this checkout, with the recipe run exactly as written:

| Scenario | Result |
|----------|-----|
| Fresh checkout, no fixture | `failing: 135` |
| Fixture (9.9.0, `src/snakemake`) + cleared sandbox | `failing: 1` |

The survivor is `Resolve implicitly imported python names > Warn about unresolved snakemake variable in run section, behaviour differs from scripts`, which the fixture does not address and which fails on `master` too.

The `grep`/`sort` recipe in the new section was checked against four real run logs — 246, 201, 145 and 3 failures — matching Gradle's own `M failed` summary line exactly in each case, scenario-outline `#5`/`#6` suffixes included.

## Reproduce

On a checkout without `testData/MockPackages3/snakemake`:

```shell
# JDK 21 is required: the pinned Gradle crashes under newer JDKs ("Type T not present").
export JAVA_HOME=/path/to/jdk-21          # e.g. `jenv prefix 21`, or an asdf/SDKMAN path
"$JAVA_HOME/bin/java" -version            # must print 21.x — see the note below

./gradlew cleanTest test 2>&1 | tee /tmp/before.log
grep -E "tests completed" /tmp/before.log   # -> 3419 tests completed, 135 failed
```

Verify `JAVA_HOME` rather than using `/usr/libexec/java_home -v 21`: that flag treats 21 as a *minimum*, so with no JDK 21 installed it returns a newer JDK and exits 0, and Gradle then dies with `Type T not present` — a confusing failure for anyone trying to reproduce these numbers.

All 135 use the unversioned `Given a snakemake project` step. By feature, the biggest groups are *Resolve implicitly imported python names* (57), the `snakemake_api.yaml` FQN check (24) and *Resolve for section names* (12). Applying the recipe in step 2 and re-running takes it to 1.

## What it deliberately does not do

- **No product code, and no new tooling.** Just the documentation fix. An earlier revision proposed a `scripts/extract_failures.py` helper and a `docs/` write-up; both are gone — see the review history below.
- **No CI workflow.** A clean runner has no prior sandbox, so a job that provisions the fixture before `./gradlew test` would pass with no cache-clear. Offered if you want it; not included here.
- **It does not make a bare `git clone && ./gradlew test` green.** Doing that means either vendoring a trimmed `MockPackages3/snakemake` fixture or automating the clone in the build. The directory appears to be gitignored deliberately (the full source is large), so that is a maintainer's call rather than something to slip into a docs fix.

## Related

- **[#572](https://github.com/JetBrains-Research/snakecharm/pull/572) (merged)** — separate issue, independent PR: that one fixed a *build* failure (`:buildWrappersBundle` dies on a hardcoded absolute path, [#571](https://github.com/JetBrains-Research/snakecharm/issues/571)); this one fixes a *test* failure. Now that it has landed, the reproduce recipe above needs no `-PsnakemakeWrappersRepoPath=…` workaround; you would only hit that on a checkout predating it.
- **[#580](https://github.com/JetBrains-Research/snakecharm/issues/580)** — snakemake is at 9.25.1 while `defaultVersion` is 9.9.0. Bumping it is separate work (16 minor releases of `introduced`/`deprecated`/`removed` entries, and probably new `MockPackages3_smk_*` mocks), but it gets easier after this PR: the recipe reads the version out of `snakemake_api.yaml`, so the fixture instructions follow the bump automatically.

<details>
<summary><b>History</b> — two reversed diagnoses and a review round that removed two thirds of the PR. Kept for anyone tracing why this ended up as a one-file change; every durable conclusion is in <code>DEVELOPER.md</code> above.</summary>

**The diagnosis flipped twice.** The first pass concluded the fixture *was* the cause, the second concluded it was not (disproved four ways), and the third established that it is — the middle round had been testing against a sandbox whose stale VFS made a correctly-provisioned fixture look inert. That is precisely the "zero effect" trap now documented in step 2, and it is why that gotcha is written as an unconditional part of the fix rather than a footnote.

**The verified number was 135 → 0 before it was 135 → 1.** The earlier figure came from a run that had not isolated the one genuine pre-existing failure. Issue #575 carried the wrong number for a while and has been corrected.

**A `scripts/extract_failures.py` helper and a `docs/missing-snakemake-fixture/` record were proposed, then removed** following [review](https://github.com/JetBrains-Research/snakecharm/pull/574#issuecomment-5383273326). The record froze a `135 failing` count that drifts as tests are added, and duplicated a diagnosis that belongs in this description and in #575. The script turned out to be a wrapper around what the Gradle console already prints: the one-liner now in `DEVELOPER.md` reproduces its output exactly across four archived runs, and the triage lists the 2026.1/2026.2 port work was actually built on had been produced that way, by grepping the console log, before the script existed.

</details>


iromeo:
Hi, I find your update to DEVELOPER.md very reasonable and useful. Thanks.  As for the rest changes:

> `scripts/README.md` 

For now, it is not mentioned in DEVELOPER.md and AGENTS.md, so this README and script are not discoverable by contributors. I think that it would be better to mention the `scripts/extract_failures.py ` in DEVELOPER.md and remove the README from the scripts folder. 

> so the practical way to see what failed is to parse the JUnit XML that `./gradlew test` leaves in `build/test-results/test/`.

For me, `./gradlew test` shows the number of failed/passed tests in the terminal output. For the rest, I use info/stats from TeamCity. I agree that the suggested script could be useful for getting a brief summary of the latest tests run when the terminal output from the previous `./gradlew test` is not available. Or if someone wants to easily compare the new run with the previous one (example from the readme). I suggest explaining more detailed `so the practical way to see what failed is to parse the JUnit XML `, otherwise there is confusion, because Gradle also shows test execution stats in the standard output

> docs/missing-snakemake-fixture

Here is the description of one setup issue that is fixed in this PR. I don't think it's worth creating a docs folder for this particular issue. Also, the magic number ` **135 failing Cucumber scenarios** ` will be changed after more tests have been implemented. I suggest just leaving the case description in the PR or ISSUE description, so it can be found through search engines, and do not introduce an extra documentation layer.

gaurav:
@iromeo Thanks for the feedback! I'll work on updating this PR sometime this week.

gaurav:
> Hi, I find your update to DEVELOPER.md very reasonable and useful. Thanks.

Thank you!

> As for the rest changes:
> 
> > `scripts/README.md`
> 
> For now, it is not mentioned in DEVELOPER.md and AGENTS.md, so this README and script are not discoverable by contributors. I think that it would be better to mention the `scripts/extract_failures.py ` in DEVELOPER.md and remove the README from the scripts folder.
>
> > so the practical way to see what failed is to parse the JUnit XML that `./gradlew test` leaves in `build/test-results/test/`.
> 
> For me, `./gradlew test` shows the number of failed/passed tests in the terminal output. For the rest, I use info/stats from TeamCity. I agree that the suggested script could be useful for getting a brief summary of the latest tests run when the terminal output from the previous `./gradlew test` is not available. Or if someone wants to easily compare the new run with the previous one (example from the readme). I suggest explaining more detailed `so the practical way to see what failed is to parse the JUnit XML `, otherwise there is confusion, because Gradle also shows test execution stats in the standard output

I interrogated Claude as to whether this script was actually useful, and it concluded that it was not: `./gradlew test` provides the test failure information, and it's pretty easy to extract the list of failing tests from it using tee + sed. So I've removed that script and its README, and I've added the tee+sed command to DEVELOPER.md in this PR (and to AGENTS.md in #570).

> > docs/missing-snakemake-fixture
> 
> Here is the description of one setup issue that is fixed in this PR. I don't think it's worth creating a docs folder for this particular issue. Also, the magic number `**135 failing Cucumber scenarios**` will be changed after more tests have been implemented. I suggest just leaving the case description in the PR or ISSUE description, so it can be found through search engines, and do not introduce an extra documentation layer.

Good point. This is now gone as well, so this PR is now down to only the DEVELOPER.md changes. I asked Claude to re-review those changes, and it found some failure cases it hadn't accounted for, which have been updated now.

Thanks for the feedback! Please let me know if you need any other additional changes.

PR #576, opened 2026-08-11
Fail the test, not the JVM, when the test wrappers bundle is missing

gaurav:
## What & why

When the test wrappers bundle is missing, `SmkWrapperStorage.loadBundledWrappersTestsMode` prints a hint to stderr and then calls `exitProcess(1)`. That kills the Gradle test worker in the middle of the run, so what actually ends the session is the JVM vanishing — the helpful part («generate it with `buildTestWrappersBundle`») is on the console, but the thing that stops the build is a worker-crash message, and nothing about it reaches the test report.

This is easy to hit: `build/` is not checked in, so any developer whose `build/bundledWrappers/smk-wrapper-storage.test.cbor` is absent — a fresh checkout where the task hasn't run, a `clean`, or a run configuration that skips `buildTestWrappersBundle` — gets a dead worker rather than a named cause.

## The change

Throw instead of exiting. The suite keeps running, the affected scenarios fail normally, and the message names both the missing file and the task that generates it:

```
Missing test wrappers bundle: '<path>/build/bundledWrappers/smk-wrapper-storage.test.cbor'.
Generate it using the `buildTestWrappersBundle` gradle task.
```

The now-unused `kotlin.system.exitProcess` import is removed. Nothing else changes — this is test-mode-only code (`loadBundledWrappersTestsMode` is reached only under `ApplicationManager.getApplication().isUnitTestMode`), so no production path is affected.

## Verification

Not just compiled — the failure path was exercised directly. I generated the bundle, moved it aside, tagged a single wrapper feature `@here`, and ran with `-x buildTestWrappersBundle` so it could not be regenerated:

```
198 tests completed, 19 failed
AllCucumberFeaturesTest > Completion for wrapper name... FAILED
    Caused by: java.lang.IllegalStateException at SmkWrapperStorage.kt:190
```

The run terminated normally and the message appears in `build/test-results/test/*.xml`. Previously the same conditions took the worker down instead.

## Scope

Independent of #572 and #574 — it touches neither the wrappers-bundle build logic nor the fixture docs, and is branched from `master` so it can be reviewed and merged on its own.

### DIFF
--- a/src/main/kotlin/com/jetbrains/snakecharm/codeInsight/completion/wrapper/SmkWrapperStorage.kt
+++ b/src/main/kotlin/com/jetbrains/snakecharm/codeInsight/completion/wrapper/SmkWrapperStorage.kt
@@ -22,7 +22,6 @@ import java.nio.file.Files
 import java.nio.file.Path
 import java.util.*
 import kotlin.io.path.exists
-import kotlin.system.exitProcess
 
 class SmkWrapperStorage(val project: Project) : Disposable {
     var version = ""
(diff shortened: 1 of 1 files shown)


iromeo:
For me, the error doesn't happen in practice; I run tests using Gradle, but I agree that your version fix is valid and useful. Thx!

PR #577, opened 2026-08-12
Port SnakeCharm to PyCharm 2026.2; the plugin now requires 2026.2+ (build 262)

### LINKED ISSUE #578: Implicit snakemake symbols do not resolve on PyCharm 2026.2
Found while porting to 2026.2 ([#577](https://github.com/JetBrains-Research/snakecharm/pull/577)). Filing separately because it is the last substantial unknown on that branch and may well be a plugin bug rather than a port artifact.

## Symptom

On 2026.2, **58 of the 76 remaining cucumber failures are references resolving to nothing**, 43 of them literally `java.lang.AssertionError: expected:<1> but was:<0>`:

| feature | failures |
|---|---|
| `Ensures that fqn used in snakemake_api.yaml corresponds to actual resolved reference fqn` | 24 |
| `Resolve implicitly imported python names` | 23 |
| `Resolve for section names in rules and checkpoints` | 11 |

The affected names are the snakemake API symbols — `temp`, `protected`, `expand`, `collect`, `lookup`, `branch`, `ancient`, `directory`, … i.e. exactly what `SmkImplicitPySymbolsProvider` supplies.

## Why it looks like a real cause rather than fallout

The same 43-failure count was **unchanged** across three independent fixes on that branch (kotlinx-serialization ABI skew 246→145, daemon-restart-during-highlighting 145→82, an inspection rename 82→76). Whatever this is, it is orthogonal to all three.

**It is not the test fixture.** With the identical `testData/MockPackages3/snakemake` fixture provisioned (snakemake v9.9.0, matching `snakemake_api.yaml`'s `defaultVersion`), the 2026.1 branch is **3245/3248** and none of these fail there.

## Working hypothesis

`SmkImplicitPySymbolsProvider` is not populating (or is populating and then being invalidated) on 2026.2 even with the snakemake library attached to the mock SDK. Note this is the same component behind the daemon-restart problem already fixed in #577, so the area is already implicated.

Possibly related: #533 (rewrite `onChange` to drop `SlowOperations`).

## Reproducing

```shell
# fixture must be provisioned first -- see DEVELOPER.md / #574
SNAKECHARM_TEST_HEAP=8g JAVA_HOME=$(jenv prefix 25) ./gradlew test --tests "features.AllCucumberFeaturesTest"
```

Read results from `build/test-results/test/*.xml` and cluster by `<failure message>` — the Gradle console log contains only exception class names, and clustering by feature makes unrelated causes look like one problem.

Full engineering log: `PORTING.md`, section "2026.2 — build 262".

gaurav:
PyCharm 2026.2 (build 262) compiles to Java 25, carries Kotlin metadata 2.4, and has removed `PyAnnotator` and the `ReturnAnnotator` extension point. The plugin #570 produces is capped at `pluginUntilBuild = 261.*` and does not load on it at all. This branch carries that port forward: three toolchain baselines that each block the build independently (Kotlin 2.3.21, JDK 25, `intellij-platform-gradle-plugin` 2.18.1), the annotator rewrite those removals force, and the one bug a user would notice — 2026.2 invalidates library PSI on every roots change, so every implicitly imported snakemake name (`expand`, `protected`, `lookup`, …) stopped resolving after any change to the project roots, with nothing rebuilding the implicit-symbol cache in time.

It requires 2026.2+ rather than declaring a range: the source binds 2026.2-only APIs, so one binary cannot also serve 2026.1.

**Stacked on #570, which GitHub cannot show.** `update-for-intellij-2026.1` lives only on the fork, so the base cannot be retargeted and the diff below is *both* ports — 46 files, roughly two thirds of the insertions #570's. Review #570 first — it is green and near-landable; what is new here is the [2026.1…2026.2 diff](https://github.com/gaurav/snakecharm/compare/update-for-intellij-2026.1...update-for-intellij-2026.2), around 24 files. Draft because it cannot merge ahead of #570 and is awaiting a human review pass, not because anything here is unfinished.

Fixes #578.

## The calls worth overruling

- **Requiring 2026.2 rather than widening the range.** Advertising compatibility the binary cannot honour is what #569 was rejected for, so `pluginSinceBuild` moves to 262 and 2026.1 users stay on #570.
- **The `return`-outside-a-function false positive is suppressed, not re-implemented.** The platform folded that check into its final `PySyntaxAnnotator`, so rather than own the check, `SmkReturnHighlightInfoFilter` drops exactly one `HighlightInfo` (Snakemake file, error severity, that bundle message, a `run:`/python-block ancestor). `return_annotator.feature` guards the narrowness from both sides — the suppression, and that a stray `yield` is still reported.
- **The symbol-cache rebuild is gated on `hasDeadPsi()`.** Rebuilding on every `rootsChanged` also clears the failures, but repopulates symbols at the moments the settings listeners deliberately cleared them, breaking `Check different SDK settings`. Dead PSI is the precise condition.
- **The suite caches project descriptors, not the SDKs they build.** Memoising the SDK is the obvious next step and does not work: the fixture disposes it with its project. Tracked as #592.

## What's here

The detail is in the repo, not here: [`PORTING.md`](https://github.com/gaurav/snakecharm/blob/update-for-intellij-2026.2/PORTING.md) → "2026.2 — build 262" is the working log — each cause, its measured effect, and the avenues tried and rejected. The rules that outlive this port went into [`AGENTS.md`](https://github.com/gaurav/snakecharm/blob/update-for-intellij-2026.2/AGENTS.md) — among them: pin anything the platform both bundles and generates code against (#587), and expect a *renamed* inspection to fail loudly where one whose check merely *moved* fails silently, in the "expect no warning" direction.

Five independent causes took the suite from 246 failing to green: a kotlinx-serialization ABI skew that `TestLoggerFactory` turned into 101 unrelated failures, a daemon restart from inside a highlighting pass, a renamed inspection, the implicit-symbol cache above, and a "disabled framework" step that never disabled anything and had been green for the wrong reason for years.

## Deliberately not here

2026.1 support, as above. The testing guide this branch keeps growing moves out of `AGENTS.md` into `docs/testing.md` in #595, stacked on top of this. Follow-up is tracked rather than done here: #583 wants measurement before a patch, #582 is a regression test for the cache rebuild, and #588 a test for the Gradle wiring that has now failed silently twice.

## Before merging

- [ ] Once #570 merges, drop the stacked-on-#570 paragraph above — the diff becomes this port alone, and the compare link stops being the useful one.

Nothing blocking. The suite is green at its full count on **both 2026.2 patch releases**: `3420 tests, 0 failures, 0 errors, 0 skipped` across 125 suites, end to end on PY-2026.2.1 (`23097522`) and again on PY-2026.2.2 (`22a94c7c`), JDK 25, `SNAKECHARM_TEST_HEAP=8g`, ~27m warm with no swap; everything on the branch since is documentation, plus one print-task change. `platformVersion` tracks the newer patch: 2026.2.2 is build `262.10315.174`, so the compatibility range already covered it, nothing in the manifest moved, and the two versions pinned *to* the platform are unchanged in it, so #587's skew cannot reappear. The count comes from `build/test-results/test/*.xml` rather than the exit code, because an all-green Gradle run prints no `N tests completed` line (#594).


gaurav:
@iromeo This PR is also ready for your review! As before, I can't stack PRs from a fork, so this includes all the changes for upstream PRs, including #570. So you'll want to review those before getting to this one so that this PR reduces to only the changes needed to support PyCharm 2026.2. Claude was able to get all the tests to run green on PyCharm 2026.2.1 and 2026.2.2 on my computer, but please let me know if they don't work on yours or if Claude introduced extra/useless/incorrect code -- nothing jumped out at me, but my Kotlin skills are not strong.

tgilon:
Thank you @gaurav for this PR! I've built the plugin locally and it seems to work with PyCharm 2.2.

<details>

Steps to build the plugin locally

```bash
git clone https://github.com/JetBrains-Research/snakecharm.git
cd snakecharm
git fetch origin pull/577/head:pr-577
git checkout pr-577

./gradlew clean buildPlugin

ls build/distributions/   # snakecharm-2026.2.0-eap.SNAPSHOT.zip
```
Then in the IDE: Settings → Plugins → gear → Install Plugin from Disk…

</details>

